PR #10, opened 2022-02-10
Add Search Clear Event

nmanu1:
Add event type for SEARCH_CLEAR_BUTTON and export AnalyticsService.

J=SLAP-1820
TEST=manual

Check that event fires correctly from test-site.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,3 +19,4 @@ export function provideAnalytics(config: AnalyticsConfig): AnalyticsReporter {
 }
 
 export * from './models';
+export { AnalyticsReporter };
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,7 @@
 import { AnalyticsConfig } from './models/AnalyticsConfig';
 import { AnalyticsReporter } from './infra/AnalyticsReporter';
 import { HttpRequester } from './infra/HttpRequester';
+import { AnalyticsService } from './services/AnalyticsService';
 
 /**
  * The entrypoint to the answers-analytics library.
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -18,4 +18,5 @@ export function provideAnalytics(config: AnalyticsConfig): AnalyticsReporter {
   return new AnalyticsReporter(config, httpRequester);
 }
 
-export * from './models';
\ No newline at end of file
+export * from './models';
+export { AnalyticsReporter };
```
yen-tt, line 22: do we need to export AnalyticsReporter if user can invoke provideAnalytics?

nmanu1: yes, we need it to add typing when defining context for it and for what's returned by a `useAnalytics` hook (see use in the component library [PR](https://github.com/yext/answers-react-components/pull/17))

oshi97: Does exporting AnalyticsService work instead?

oshi97: it might even make sense for provideAnalytics to have return type `AnalyticsService` instead

oshi97: if I'm being perfectly honest I'm not sure what the `AnalyticsService` interface existence gets us in the first place besides letting us rename `AnalyticsReporter` and add arbitrary non-private properties, but this is probably just fussing over minutiae

nmanu1: yeah, `AnalyticsService` works. I can use that instead for `provideAnalytics`

oshi97 (approved):

PR #12, opened 2022-02-11
add autocomplete analytics event

yen-tt:
add autocomplete analytics event for when user select an autocomplete suggestion

J=SLAP-1821
TEST=manual

use analytics test-site to fire an `AUTO_COMPLETE_SELECTION` event and see 200 status.

### DIFF
--- a/src/models/AnalyticsEventType.ts
+++ b/src/models/AnalyticsEventType.ts
@@ -13,6 +13,8 @@ export enum AnalyticsEventType {
   VoiceStop = 'VOICE_STOP',
   SearchBarImpression = 'SEARCH_BAR_IMPRESSION',
   SearchClearButton = 'SEARCH_CLEAR_BUTTON',
+  Paginate = 'PAGINATE',
+  AutocompleteSelection = 'AUTO_COMPLETE_SELECTION',
 
   // CTA event types
   CtaClick = 'CTA_CLICK',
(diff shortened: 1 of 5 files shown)


src/models/AnalyticsEventType.ts
```diff
@@ -28,5 +28,8 @@ export enum AnalyticsEventType {
   Rsvp = 'RSVP',
 
   // Pagination event types
-  Paginate = 'PAGINATE'
+  Paginate = 'PAGINATE',
```
nmanu1, line 31: nit: I think the CTA event types are listed separately because there are a lot of them, but since the other events just have one or two types, we've just grouped them together at the top, which seems a little more concise than separating by event. but I don't have a strong preference if you'd rather keep them separate

src/models/events/AutocompleteEvent.ts
```diff
@@ -0,0 +1,9 @@
+import { AnalyticsEventType } from '../AnalyticsEventType';
+import { EnumOrLiteral } from '../utils';
+
+/** Event for autocomplete selection. */
+export interface AutocompleteEvent {
```
nmanu1, line 5: I think in the SDK we send `queryId` with this (even if it's not required)?

src/models/AnalyticsEventType.ts
yen-tt: sure

src/models/events/AutocompleteEvent.ts
yen-tt: sure thing, updated to include it optionally. I believe visitor field is also included in SDK analytics events from the few events I just tested. Made a separate item for that

PR #18, opened 2022-03-10
Add API Extractor

cea2aj:
Add API extractor, add missing exports and generate docs

API extractor provides us with an API report, generated docs, and a type bundle. Also add the GitHub action which automatically updates the docs. A future PR will include updating the documentation for missing descriptions, and another PR will include the README.

J=SLAP-1946
TEST=manual

Run a build and inspect generated docs, API report and type bundle


src/infra/AnalyticsReporter.ts
```diff
@@ -8,8 +8,12 @@ import { Visitor } from '../models/Visitor';
 
 const defaultDomain = 'https://answers.yext-pixel.com';
 
+/**
+ * @public
+ */
 export class AnalyticsReporter implements AnalyticsService {
   private _visitor: Visitor | undefined;
+
   constructor(private config: AnalyticsConfig, private httpRequesterService: HttpRequesterService) {
```
nmanu1, line 17: Should this constructor be marked as internal?

src/index.ts
```diff
@@ -20,4 +20,5 @@ export function provideAnalytics(config: AnalyticsConfig): AnalyticsService {
 }
 
 export * from './models';
-export { AnalyticsService };
\ No newline at end of file
+export * from './services';
+export { AnalyticsReporter };
```
nmanu1, line 24: Do we need to export `AnalyticsReporter`? I thought `AnalyticsService` would be sufficient?

src/infra/AnalyticsReporter.ts
```diff
@@ -40,10 +44,10 @@ export class AnalyticsReporter implements AnalyticsService {
 
   /**
    * Formats the event data for the api which includes adapting verticalKey to verticalConfigId
```
nmanu1, line 46: nit: missing period at the end here and in a few other files' docs (e.g. AnalyticsEventType, AnalyticsResponse, HttpRequesterService)

src/services/index.ts
```diff
@@ -0,0 +1,2 @@
+export { AnalyticsService } from './AnalyticsService';
+export { HttpRequesterService } from './HttpRequesterService';
```
nmanu1, line 2: Does HttpRequesterService need to be exported? I didn't think it would be accessed by users of the library

src/infra/AnalyticsReporter.ts
```diff
@@ -8,8 +8,12 @@ import { Visitor } from '../models/Visitor';
 
 const defaultDomain = 'https://answers.yext-pixel.com';
 
+/**
+ * @public
```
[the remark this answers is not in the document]
cea2aj: Sure, if there isn't a use case for it, we can keep it private for now and always make it public in the future if needed

src/infra/AnalyticsReporter.ts
cea2aj: Going to make the whole component internal per Tom's comment

src/index.ts
cea2aj: we no longer need to export it since it's now internal

src/services/index.ts
cea2aj: The HttpRequesterService would need to be exported if we make `AnalyticsReporter` public, however since we are now making it internal, `HttpRequesterService` can be internal as well

package.json
```diff
@@ -5,13 +5,16 @@
...
   "files": [
     "src",
     "lib"
   ],
   "scripts": {
-    "build": "tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json",
+    "build": "rm -rf lib/** && tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json && npm run api-extractor && npm run generate-docs",
     "lint": "eslint .",
+    "api-extractor": "api-extractor run --local --verbose",
+    "generate-docs": "api-documenter markdown --input-folder temp --output-folder docs && rm -rf temp",
```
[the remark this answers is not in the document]
cea2aj: api-extractor creates the api report, the type bundle, and a [*.api.json](https://www.npmjs.com/package/@microsoft/api-extractor-model) file which contains the ApiModel. the ApiModel is a big tree of the api and everything needed to generate the documentation. api-documenter consumes this ApiModel and builds the markdown files from it. The ApiModel is strongly typed in TypeScript, so we could easily render API reference docs in a React app with our own custom styling if we wanted to.

package.json
```diff
@@ -5,13 +5,16 @@
...
   "main": "./lib/commonjs/index.js",
   "module": "./lib/esm/index.js",
+  "types": "./lib/analytics.d.ts",
   "files": [
     "src",
     "lib"
   ],
   "scripts": {
-    "build": "tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json",
+    "build": "rm -rf lib/** && tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json && npm run api-extractor && npm run generate-docs",
```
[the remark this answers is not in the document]
cea2aj: yes, it's a custom script so we need `run` here. Only a few scripts such as `test` can be ran without `run` in front of the script name.

src/infra/AnalyticsReporter.ts
cea2aj: The missing periods should be addressed in #20

tmeyer2115 (approved):

PR #21, opened 2022-03-21
use fetch instead of navigator.beacon

oshi97:
Updates requests to use the fetch API with keepalive: true
instead of navigator.beacon, so that in the future we can
set custom HTTP headers.

J=SLAP-1968
TEST=manual

use the test-site to test that promise rejections are performed
as expected


test-site/src/index.ts
```diff
@@ -6,12 +6,26 @@ const analytics = provideAnalytics({
...
+    type: 'CTA_CLICK',
+    queryId: '95751527-9db6-4859-8278-60d1c060b6c0'
+  });
+}
+
+function fireAnalyticsEvent(event) {
+  analytics
+    .report(event)
+    .then(() => console.log('success!'))
+    .catch((err => console.error(err)));
```
yen-tt, line 30: nit: extra parentheses in catch

test-site/webpack.config.js
```diff
@@ -24,13 +23,6 @@ module.exports = [{
   resolve: {
     extensions: [ '.tsx', '.ts', '.js' ],
   },
-  plugins: [
-    new CopyPlugin({
-      patterns: [
-        { from: 'src/index.html' },
-      ],
```
yen-tt, line 31: Without this, we wouldn't be able to view the stuff in src/index.html since the test-site `npm run serve` command serves the dist folder. Maybe we should keep it?

oshi97: oh I see, sometimes (like today) I just open the file directly in my browser, but that's fair!

src/infra/HttpRequester.ts
```diff
@@ -1,42 +1,18 @@
...
-  _beaconPolyfill(url: string, data: string): boolean {
-    const event = window.event?.type;
-    const sync = event === 'unload' || event === 'beforeunload';
-    const xhr = ('XMLHttpRequest' in window) ? new XMLHttpRequest() : new ActiveXObject('Microsoft.XMLHTTP');
-    xhr.open('POST', url, !sync);
-    xhr.setRequestHeader('Accept', '*/*');
-    xhr.setRequestHeader('Content-Type', 'text/plain;charset=UTF-8');
-    xhr.send(data);
-    return true;
+    return fetch(url, {
```
yen-tt, line 12: fetch doesn't work for IE, do we need some sort of polyfill?

oshi97: I don't think we need to do any polyfilling here, answers-core also doesn't do any polyfilling in core but works fine in ie11. This is probably due to cross-fetch already using whatwg-fetch under the hood.
I added `import 'core-js/stable'` to the top of the test-site's index.ts and the site works fine in ie11.

oshi97: I updated HttpRequester to use window.fetch if available the same way core does it, and retested in ie11!

cea2aj:
+1

yen-tt (approved):

PR #27, opened 2022-07-07
update description to Search and add list of approved keywords

juliannzhou:
This PR adds a list of approved keywords to package.json and updates Answers to Search in description.

J=SLAP-2190
TEST=manual

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,12 +1,18 @@
 {
   "name": "@yext/analytics",
   "version": "0.2.0-beta.5",
-  "description": "An analytics library for Yext Answers",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,12 +1,18 @@
 {
   "name": "@yext/analytics",
   "version": "0.2.0-beta.5",
-  "description": "An analytics library for Yext Answers",
+  "description": "An analytics library for Yext Search",
```
nmanu1, line 4: I don't think this library is intended only for Search

PR #29, opened 2022-07-29
implement conversion tracking service

benmcginnis:
Adds Conversion Tracking to the Analytics Library, including:
* New standalone ConversionTrackingReporter
* Additions to PagesAnalyticsReporter to allow it to track conversions using the ConversionTrackingReporter as well as properly credit Listings and Pages for conversions that happen later
* Adds a CookieManager utility class that can be used by implementers to retrieve the Yext cookieId and set it if need be.


src/infra/ConversionTrackingReporter.ts
```diff
@@ -0,0 +1,75 @@
...
+    params.set(COOKIE_PARAM, event.cookieId);
+    if (event.referrer) params.set('referrer', event.referrer);
+    params.set('v', calculateSeed().toString());
+  }
+
+  /** {@inheritDoc ConversionTrackingService.trackConversion} */
+  async trackConversion(event: ConversionEvent): Promise<void> {
+    const url = new URL(`https://${DEFAULT_CONVERSION_TRACKING_DOMAIN}/${conversionEndpoint}`);
+    const params = new URLSearchParams();
+    params.set('cid', event.cid);
```
AjayBenno, line 51: A note, CID id a required parameter if we want so the server will 400 if this doesn't exist.

src/infra/ConversionTrackingReporter.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private static formatBaseEvent(event: CommonConversionData, params: URLSearchParams): void {
+    params.set(COOKIE_PARAM, event.cookieId);
+    if (event.referrer) params.set('referrer', event.referrer);
+    params.set('v', calculateSeed().toString());
+  }
+
+  /** {@inheritDoc ConversionTrackingService.trackConversion} */
+  async trackConversion(event: ConversionEvent): Promise<void> {
```
AjayBenno, line 48: Note we also need to pass a location field for the conversion event as well(its should be the same as the listings location param(currently for ytag it comes from window.location.href)

src/infra/ConversionTrackingReporter.ts
```diff
@@ -0,0 +1,75 @@
...
+    params.set('cid', event.cid);
+    if (event.cv) params.set('cv', event.cv);
+    ConversionTrackingReporter.formatBaseEvent(event, params);
+    url.search = params.toString();
+    await this.handleRequest(url.toString());
+    this.printEvent(event.cid, 'Conversion');
+  }
+
+  /** {@inheritDoc ConversionTrackingService.trackListings} */
+  async trackListings(event: ListingsClickEvent): Promise<void> {
```
AjayBenno, line 60: Curious, how is this expected to be used? I'm reading through the ytag code and the listings stuff has some special logic where it only fires once upon load.

src/infra/ConversionTrackingReporter.ts
```diff
@@ -0,0 +1,75 @@
...
+    const url = new URL(`https://${DEFAULT_CONVERSION_TRACKING_DOMAIN}/${listingsEndpoint}`);
+    const params = new URLSearchParams();
+    params.set(LISTINGS_SOURCE_PARAM, event.source);
```
AjayBenno, line 63: Also note, /listings will throw a server error if y_source is not set

src/infra/ConversionTrackingReporter.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  async handleRequest(url: string): Promise<void> {
+    const res = await this.httpRequesterService.get(url.toString());
+    if (res.status !== 200 && !(res.type === 'opaqueredirect' || res.type === 'opaque')) {
+      const errorMessage = await res.text();
+      throw new Error(errorMessage);
+    }
+  }
+
+  private static formatBaseEvent(event: CommonConversionData, params: URLSearchParams): void {
```
AjayBenno, line 41: Are we only dealing with first party cookies for sites?

AjayBenno (commented):
Left a few comments on required parameters(if that influences any preconditions we want to enforce client side) and some questions.

src/infra/ConversionTrackingReporter.ts
benmcginnis: yes, it's marked required on the ConversionEvent interface

src/infra/ConversionTrackingReporter.ts
benmcginnis: I didn't see that on the required param for the conversion event itself, just the listingsclickevent in the docs, it's not even listed as an optional parameter on the conversion event: https://www.yext.com/docmd/docs/src/com/yext/analytics/events/conversions.html

src/infra/ConversionTrackingReporter.ts
benmcginnis: So some of the borders of this library are a bit weird because it's meant to be useable server side.  I need to think about the pageView method and what it should provide implementers to tell them to remove the listings parameter from the pageurl if necessary.  Any thoughts?  I can pull the listings crediting out of the pageView method so that it's entirely on the implementer to do and add that logic into `yext/pages` (seems like that might be best).  I could also add something to utils for implementers to use to manage that state

src/infra/ConversionTrackingReporter.ts
benmcginnis: it's required by the ListingsClickEvent

src/infra/ConversionTrackingReporter.ts
benmcginnis: after discussing with @maperalta the `cookieId` is either the first or the third party cookie.  This library doesn't send anything in the headers as it doesn't have direct access to the Document, which is why it seems a bit weird.

If you look at CookieManager in `utils` that has the cookie retrieval logic from ytag and in `yext/pages` I'm using it like this:

```
  enableConversionTracking(): void {
    if (this.canTrack()) {
      this._cookieManager = new CookieManager();
      this._analyticsReporter?.setConversionTrackingEnabled(
        true,
        this._cookieManager.setAndGetYextCookie(),
      );
    }
  }

```

benmcginnis: let me also tag a WIP PR for yext/pages so you can see the implementation on the front end

benmcginnis: Here's the proposed usage: https://github.com/yext/pages/pull/151/files#diff-b1eff6c672541942b411af90724e5a14038cd68e1ef1f94333ad26c859c51dc7R80

src/infra/ConversionTrackingReporter.ts
AjayBenno: Ah yea location is a new param we added in the last few months for conversions(its not required for the event to work though). Its mainly to support our new analytics events pipeline. I'll update those docs.

src/infra/ConversionTrackingReporter.ts
AjayBenno: Oh i see the code in PagesAnalyticsReporter where it does the only fire once logic. I think we might actually be ok with what logic already exists there, it seems that on the ytag implementation refreshing the page would cause the listings event to fire again. 

Up to you if you want to move it out right now, think it should be ok.

benmcginnis: oh cool I was just looking at ytag and hadn't seen your comment and yeah I agree it doesn't seem to handle page refreshes.  I can add a TODO to handle that in the future

AjayBenno (approved):

tmeyer2115 (approved):

PR #28, opened 2022-07-15
feat(store_pagespixel): add legacy analytics support

tatimblin:
J=CR-2409

PR #30, opened 2022-08-09
Add Support for Pages Analytics

tmeyer2115:
Support was added for Pages Analytics Events and Conversion Tracking.


benmcginnis:
LGTM, it has both commits that were already reviewed in #29 and #28

PR #31, opened 2022-09-27
update queryParam from id to ids to match pages pixel documentation

willgorick:
updating the id query param to ids to match the documentation [here](https://www.yext.com/docmd/docs/src/com/yext/analytics/events/pagesevents.html#request-parameters) and to fix entity_id not associating properly with events bug


benmcginnis:
this fixes: #32

benmcginnis (approved):
LGTM

PR #33, opened 2022-09-29
update visitor field id -> idMethod to match search analytics format

willgorick:


src/infra/PagesAnalyticsReporter.ts
```diff
@@ -119,7 +119,7 @@ export class PagesAnalyticsReporter implements PagesAnalyticsService{
 
     if (this._visitor) {
       params.set(urlParamNames.VisitorId, this._visitor.id);
-      if (this._visitor.method) params.set(urlParamNames.VisitorMethod, this._visitor.method);
+      if (this._visitor.idMethod) params.set(urlParamNames.VisitorMethod, this._visitor.idMethod);
```
benmcginnis, line 122: I think instead of doing a find / replace here you should just change `urlParamNames.VisitorMethod` as that will only change the URL param that is sent: https://github.com/yext/analytics/pull/33/files#diff-fd21ce04d1b95f6bc7169267501761e512c59f87fe1b8b249b6021b3aa849790L30

That way we won't break the interface and require changes in the search and pages libraries.

I think you also probably need to change the conversion tracking and search reporters too, or at least check that they have the correct url params

benmcginnis: ah ok, thanks for the explanation on slack.  I didn't understand that the problem was the inconsistency in interface between the two libraries

benmcginnis (approved):

PR #36, opened 2022-10-05
style: rename tsconfig.base.json to tsconfig.json

tatimblin:
J=SLAP-2389
TEST=eslint,manual

ensured tests weren't included in lib/


yen-tt (commented):
could you update this to point to develop instead? we generally only merge directly to main for releases/patches!

PR #38, opened 2023-06-21
chat analytics

mdavish:
This PR introduces a `ChatAnalyticsProvider` along with `provideChatAnalytics`. This class interfaces with the new Analytics Events API, which eventually all products will use. Additionally, this PR introduces some minor VS Code settings for consistent formatting and improved developer experience.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,27 +1,31 @@
 module.exports = {
-  extends: [
-    'plugin:@typescript-eslint/recommended'
-  ],
+  extends: ['plugin:@typescript-eslint/recommended'],
   parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaVersion: 2020,
-    sourceType: 'module'
+    sourceType: 'module',
   },
   rules: {
     indent: ['error', 2, { SwitchCase: 1 }],
-    'no-trailing-spaces' : ['error'],
-    'no-multi-spaces' : ['error'],
+    'no-trailing-spaces': ['error'],
+    'no-multi-spaces': ['error'],
     quotes: ['error', 'single'],
-    'space-before-function-paren': ['error', {
-      named: 'never',
-      anonymous: 'never'
-    }],
+    'space-before-function-paren': [
+      'error',
+      {
+        named: 'never',
+        anonymous: 'never',
+      },
+    ],
     'keyword-spacing': ['error'],
     'quote-props': ['error', 'as-needed'],
-    'max-len': ['error', {
-      code: 110,
-      ignorePattern: '^import\\s.+\\sfrom\\s.+;$'
-    }],
+    'max-len': [
+      'error',
+      {
+        code: 110,
+        ignorePattern: '^import\\s.+\\sfrom\\s.+;$',
+      },
+    ],
     '@typescript-eslint/semi': ['error'],
     '@typescript-eslint/type-annotation-spacing': ['error'],
   },
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,14 @@
+{
+  "[typescript]": {
+    "editor.defaultFormatter": "dbaeumer.vscode-eslint",
+    "editor.codeActionsOnSave": {
+      "source.fixAll.eslint": true
+    }
+  },
+  "[javascript]": {
+    "editor.defaultFormatter": "dbaeumer.vscode-eslint",
+    "editor.codeActionsOnSave": {
+      "source.fixAll.eslint": true
+    }
+  }
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,5 +10,13 @@
     "editor.codeActionsOnSave": {
       "source.fixAll.eslint": true
     }
+  },
+  "[markdown]": {
+    "editor.formatOnSave": false,
+    "editor.wordWrap": "wordWrapColumn",
+    "editor.wordWrapColumn": 100,
+    "editor.rulers": [
+      100
+    ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -38,9 +38,13 @@
     "@typescript-eslint/parser": "^4.28.2",
     "babel-jest": "^27.0.6",
     "eslint": "^7.30.0",
+    "eslint-config-prettier": "^8.8.0",
+    "eslint-plugin-prettier": "^4.2.1",
     "generate-license-file": "^1.3.0",
     "isomorphic-fetch": "^3.0.0",
     "jest": "^27.0.6",
+    "jest-fetch-mock": "^3.0.3",
+    "prettier": "^2.8.8",
     "typescript": "^4.1.5"
   },
   "jest": {
@@ -65,6 +69,7 @@
     ]
   },
   "dependencies": {
-    "cross-fetch": "^3.1.5"
+    "cross-fetch": "^3.1.5",
+    "dotenv": "^16.3.1"
   }
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,8 +3,10 @@ import { AnalyticsReporter } from './infra/AnalyticsReporter';
 import { ConversionTrackingReporter } from './infra/ConversionTrackingReporter';
 import { SearchAnalyticsReporter } from './infra/SearchAnalyticsReporter';
 import { PagesAnalyticsReporter } from './infra/PagesAnalyticsReporter';
+import ChatAnalyticsReporter from './infra/ChatAnalyticsReporter';
 import { AnalyticsConfig, PagesAnalyticsConfig, SearchAnalyticsConfig } from './models';
 import { AnalyticsService, SearchAnalyticsService, PagesAnalyticsService, ConversionTrackingService } from './services';
+import ChatAnalyticsConfig from './models/config/ChatAnalyticsConfig';
 
 /**
  * Maintained for backwards compatibility with older versions of the library
@@ -66,6 +68,18 @@ export function providePagesAnalytics(config: PagesAnalyticsConfig): PagesAnalyt
 export function provideConversionTrackingAnalytics(debug?: boolean): ConversionTrackingService {
   const httpRequester = new HttpRequester();
   return new ConversionTrackingReporter(httpRequester, debug);
+}
+
+/**
+ * Provides a Chat Analytics service
+ *
+ * @remarks
+ * Returns a conversion tracking service instance
+ *
+ * @public
+ */
+export function provideChatAnalytics(config: ChatAnalyticsConfig): ChatAnalyticsReporter {
+  return new ChatAnalyticsReporter(config);
 }
 
 export * from './models';
--- a/src/infra/ChatAnalyticsReporter.ts
+++ b/src/infra/ChatAnalyticsReporter.ts
@@ -1,0 +1,74 @@
+import type { EventPayload } from '../models/chat';
+import type ChatAnalyticsConfig from '../models/config/ChatAnalyticsConfig';
+
+interface EventsAPIResponse {
+  id: string;
+}
+
+/**
+ * A class to report chat analytics data. Uses the provided API key, environment,
+ * and partition to determine the endpoint for reporting events.
+ */
+export default class ChatAnalyticsReporter {
+  public readonly apiKey: string;
+  public readonly env: NonNullable<ChatAnalyticsConfig['env']>;
+  public readonly partition: NonNullable<ChatAnalyticsConfig['partition']>;
+  public readonly endpoint: string;
+
+  private endpoints: Record<
+    NonNullable<ChatAnalyticsConfig['partition']>,
+    Record<NonNullable<ChatAnalyticsConfig['env']>, string>
+  > = {
+    US: {
+      PROD: 'https://www.us.yextevents.com/accounts/me/events',
+      SANDBOX: 'https://www.sbx.us.yextevents.com/accounts/me/events',
+      QA: 'https://www.qa.us.yextevents.com/accounts/me/events',
+    },
+    EU: {
+      PROD: 'https://www.eu.yextevents.com/accounts/me/events',
+      SANDBOX: 'https://www.sbx.eu.yextevents.com/accounts/me/events',
+      QA: 'https://www.qa.eu.yextevents.com/accounts/me/events',
+    }
+  };
+
+  constructor({ apiKey, env, partition }: ChatAnalyticsConfig) {
+    this.apiKey = apiKey;
+    this.env = env ?? 'PROD';
+    this.partition = partition ?? 'US';
+    this.endpoint = this.endpoints[this.partition][this.env];
+  }
+
+  /**
+   * Report an event to the chat analytics API.
+   * @param event The event to report.
+   * @returns A promise that resolves to the response from the API.
+   * @example
+   * reporter.report({
+   *  action: 'DIRECTIONS',
+   *   chat: {
+   *   botId: 'davish-playground',
+   * },
+   * });
+   * // Response
+   * {
+   *  id: '12345',
+   * }
+  */
+  public async report(event: EventPayload): Promise<EventsAPIResponse> {
+    const res = await fetch(this.endpoint, {
+      method: 'POST',
+      headers: {
+        Authorization: `KEY ${this.apiKey}`,
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(event),
+    });
+    if (!res.ok) {
+      throw new Error(
+        `Events API responded with ${res.status}. ${res.statusText}`
+      );
+    }
+    const resJson = await res.json();
+    return resJson as EventsAPIResponse;
+  }
+}
--- a/src/models/chat/index.ts
+++ b/src/models/chat/index.ts
@@ -1,0 +1,115 @@
+/**
+ * The action types accepted by the Analytics Events API.
+ *
+ * @public
+ */
+export const ACTIONS = [
+  'ADD_TO_CART',
+  'ALL_TAB_NAVIGATION',
+  'APPLY',
+  'AUTO_COMPLETE_SELECTION',
+  'BACKWARD_PAGINATE',
+  'BOOK',
+  'BRAND_ICON',
+  'CALL_TO_ACTION',
+  'CASE_START',
+  'CASE_SUBMITTED',
+  'CHAT_IMPRESSION',
+  'CHAT_LINK_CLICK',
+  'CHAT_RESPONSE',
+  'COLLAPSE',
+  'DIRECTIONS',
+  'EVENT',
+  'EXPAND',
+  'FEATURED_MESSAGE',
+  'FILTERING_WITHIN_SECTION',
+  'FORWARD_PAGINATE',
+  'HEADER_LINKS',
+  'ITEM_IN_LIST',
+  'MAP_CARD',
+  'MAP_PIN',
+  'MENU',
+  'MESSAGE',
+  'ORDER',
+  'PAGINATE',
+  'PHONE',
+  'POST',
+  'PRESET_PROMPT',
+  'PRODUCT',
+  'PROFILE',
+  'QUESTION_FOCUS',
+  'QUESTION_SUBMIT',
+  'REMOVED_FILTER',
+  'REVIEW',
+  'SCROLL_TO_BOTTOM_OF_PAGE',
+  'SEARCH_BAR_IMPRESSION',
+  'SEARCH_CLEAR_BUTTON',
+  'THUMBS_DOWN',
+  'THUMBS_UP',
+  'TICKET_URL',
+  'TITLE',
+  'VERTICAL_TAB_NAVIGATION',
+  'VERTICAL_VIEW_ALL',
+  'VOICE_START',
+  'VOICE_STOP',
+  'WEBSITE',
+] as const;
+
+/**
+ * The action types accepted by the Analytics Events API.
+ *
+ * @public
+ */
+export type Action = (typeof ACTIONS)[number];
+
+/**
+ * The payload accepted by the Analytics Events API.
+ *
+ * @public
+ */
+export interface EventPayload {
+  action: Action;
+  sessionId?: string;
+  pageUrl?: string;
+  destinationUrl?: string;
+  label?: string;
+  locale?: string;
+  timestamp?: string;
+  bot?: string; // Whether the event is from a bot
+  browserAgent?: {
+    browser?: string;
+    browserVersion?: string;
+    os?: string;
+    osVersion?: string;
+    device?: string;
+    deviceClass?: string;
+    userAgent?: string;
+  };
+  clientSdk?: Record<string, string>;
+  count?: number;
+  customTags?: Record<string, string>;
+  customValues?: Record<string, number>;
+  chat?: ChatDomainProperties;
+  entity?:
+    | {
+        entityId: string;
+      }
+    | {
+        entityUid: number;
+      };
+  ip?: {
+    address: string;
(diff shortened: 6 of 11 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,22 @@
+{
```
benmcginnis, line 1: can you maybe git ignore this?

.eslintrc.js
```diff
@@ -1,29 +1,33 @@
 module.exports = {
-  extends: [
-    'plugin:@typescript-eslint/recommended'
-  ],
+  extends: ['plugin:@typescript-eslint/recommended'],
```
benmcginnis, line 2: can you reset this file, it doesn't look like you changed anything

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -0,0 +1,74 @@
...
+
+interface EventsAPIResponse {
+  id: string;
+}
+
+/**
+ * A class to report chat analytics data. Uses the provided API key, environment,
+ * and partition to determine the endpoint for reporting events.
+ */
+export default class ChatAnalyticsReporter {
```
benmcginnis, line 12: I think this library generally avoids default imports

src/index.ts
```diff
@@ -3,8 +3,10 @@ import { AnalyticsReporter } from './infra/AnalyticsReporter';
 import { ConversionTrackingReporter } from './infra/ConversionTrackingReporter';
 import { SearchAnalyticsReporter } from './infra/SearchAnalyticsReporter';
 import { PagesAnalyticsReporter } from './infra/PagesAnalyticsReporter';
+import ChatAnalyticsReporter from './infra/ChatAnalyticsReporter';
 import { AnalyticsConfig, PagesAnalyticsConfig, SearchAnalyticsConfig } from './models';
 import { AnalyticsService, SearchAnalyticsService, PagesAnalyticsService, ConversionTrackingService } from './services';
+import ChatAnalyticsConfig from './models/config/ChatAnalyticsConfig';
```
benmcginnis, line 9: can you switch to a non-default import for these please?

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -0,0 +1,74 @@
...
+  public readonly endpoint: string;
+
+  private endpoints: Record<
+    NonNullable<ChatAnalyticsConfig['partition']>,
+    Record<NonNullable<ChatAnalyticsConfig['env']>, string>
+  > = {
+    US: {
+      PROD: 'https://www.us.yextevents.com/accounts/me/events',
+      SANDBOX: 'https://www.sbx.us.yextevents.com/accounts/me/events',
+      QA: 'https://www.qa.us.yextevents.com/accounts/me/events',
```
benmcginnis, line 25: I don't think we publish our QA/DEV urls generally

benmcginnis: In search we export this endpoints object: https://github.com/yext/search-core/blob/master/src/models/core/Endpoints.ts

benmcginnis: and we build the object here: https://github.com/yext/search-core/blob/master/src/provideEndpoints.ts#L13

benmcginnis: which lets you pass in other environments that are un-documented

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -0,0 +1,74 @@
...
+   * // Response
+   * {
+   *  id: '12345',
+   * }
+  */
+  public async report(event: EventPayload): Promise<EventsAPIResponse> {
+    const res = await fetch(this.endpoint, {
+      method: 'POST',
+      headers: {
+        Authorization: `KEY ${this.apiKey}`,
```
benmcginnis, line 61: this is new, we don't use the key as a query param?

src/models/chat/index.ts
```diff
@@ -0,0 +1,115 @@
+/**
+ * The action types accepted by the Analytics Events API.
+ *
+ * @public
+ */
+export const ACTIONS = [
+  'ADD_TO_CART',
```
benmcginnis, line 7: can you make these:
```
ADD_TO_CART: 'ADD_TO_CART',
```

So they can be references like `Actions.ADD_TO_CART`?

src/models/chat/index.ts
```diff
@@ -0,0 +1,115 @@
+/**
+ * The action types accepted by the Analytics Events API.
+ *
+ * @public
+ */
+export const ACTIONS = [
```
benmcginnis, line 6: I think this can just be `Actions`

src/models/chat/index.ts
```diff
@@ -0,0 +1,115 @@
...
+export interface EventPayload {
+  action: Action;
+  sessionId?: string;
+  pageUrl?: string;
+  destinationUrl?: string;
+  label?: string;
+  locale?: string;
+  timestamp?: string;
+  bot?: string; // Whether the event is from a bot
+  browserAgent?: {
```
benmcginnis, line 79: You're going to move setting this stuff into the chat sdks right? so people don't have to build this massive object manually?

benmcginnis (commented):
Overall LG

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -0,0 +1,74 @@
+import type { EventPayload } from '../models/chat';
+import type ChatAnalyticsConfig from '../models/config/ChatAnalyticsConfig';
+
+interface EventsAPIResponse {
+  id: string;
```
AjayBenno, line 5: it's also possible for an error field to be returned in this response if its not a success.

`{
    "id": "01H3FNTWFVES5KF3J29V5V33PW",
    "errors": [...]
}`

src/infra/ChatAnalyticsReporter.ts
AjayBenno: Yea for the events api we're accepting the API key via the Authorization Header or via the requests body in an authorization field.

src/models/chat/index.ts
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Analytics event properties specific to Yext Chat.
+ *
+ * @public
+ */
+export interface ChatDomainProperties {
+  botId: string;
+  conversationId?: string;
+  sessionId?: string;
```
AjayBenno, line 114: Session Id is currently not a property of the chat domain. We're also missing responseId.  

Current chat domain looks like https://gerrit.yext.com/plugins/gitiles/alpha/+/master/src/com/yext/analytics/events/resources/analytics-event-v1.json#278

benmcginnis: How would one register a sessions start for chat?

.eslintrc.js
mdavish: Yep will do

.vscode/settings.json
mdavish: I was finding it pretty useful and thought others might to, since your global settings might conflict with this particular project. (E.g. my settings enforce double quotes in Typescript by default, whereas this repo does single quotes.)

src/infra/ChatAnalyticsReporter.ts
mdavish: Yes it appears so. I change that.

src/index.ts
mdavish: Yep!

src/models/chat/index.ts
mdavish: Yep that's the plan

src/models/chat/index.ts
```diff
@@ -0,0 +1,115 @@
...
+ * @public
+ */
+export interface EventPayload {
+  action: Action;
+  sessionId?: string;
+  pageUrl?: string;
+  destinationUrl?: string;
+  label?: string;
+  locale?: string;
+  timestamp?: string;
```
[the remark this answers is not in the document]
mdavish: We could use Javascript's Date() class and then automatically convert it to a string of the before sending it to the API. Would that be better?

src/infra/ChatAnalyticsReporter.ts
mdavish: Ok I can get a rid of the QA URL. I can also add the ability to manually specify a `domain`. But I think we should keep the current default behavior of accepting an `env` and a `partition` and then figuring out the URL for you, because otherwise it's very annoying to figure out what domain you're actually looking for. Does that work?

src/models/chat/index.ts
mdavish: I can do that. Just out of curiosity, what's the benefit of doing it that way?

src/models/chat/index.ts
mdavish: Yep that works.

src/infra/ChatAnalyticsReporter.ts
mdavish: Are `errors` and `id` mutually exclusive? Or can there be both?

If there are `errors`, what will the status code be? Because right now we throw an error when if the status is not 2xx. 

We could potentially change this behavior so the user can access data about the error. Let me know what you think.

src/models/chat/index.ts
mdavish: Yeah good call, this is a mistake. Also @AjayBenno - should we add the pages domain as well? A chat analytics event can take place on a Yext Page, after all.

@benmcginnis - right now you will have to do this manually, but at some point soon (perhaps in this pull request) we should add a method to this SDK for `startSession` and `getSession`, to standardize our approach to this.

.vscode/settings.json
mdavish: Do we have VS Code settings in any of our other repos? If not I can get rid of it, but it might be a nice convenience for developers who maintain different personal settings.

src/models/chat/index.ts
mdavish: I made this change and now you have to import the `Actions` enum in order to use `.report()`, e.g. `analytics.report({ action: Actions.ADD_TO_CARD })`

To me this feels a lot more annoying than just using the string literal type, for which I get nice intellisense. Is there a benefit to doing this with enums?

src/models/chat/index.ts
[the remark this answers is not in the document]
benmcginnis: That would probably be more convenient and would ensure that the event is sent in the correct format for the API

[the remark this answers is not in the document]
mdavish: Yeah agreed, I'll go ahead and change that

.vscode/settings.json
benmcginnis: Honestly I don't know, I don't think we have them. Usually editors will derive the settings from eslint settings rather that requiring their own settings file.  @tmeyer2115 ?

src/models/chat/index.ts
benmcginnis: So the const array isn't a string literally type. I think you could also do `type Actions 'FOO|BAR` which might be more what you want?  I do the enum style in javascript mostly, sometimes I forget about the niceties in typescript.

package.json
```diff
@@ -38,9 +38,13 @@
     "@typescript-eslint/parser": "^4.28.2",
     "babel-jest": "^27.0.6",
     "eslint": "^7.30.0",
+    "eslint-config-prettier": "^8.8.0",
+    "eslint-plugin-prettier": "^4.2.1",
```
yen-tt, line 42: I don't think these eslint plugins and extensions are being used. Can we remove these deps?

.vscode/settings.json
yen-tt: Tom is out this week -- We don't check in personal settings like these into our repos, any rules we want to enforce are usually through eslint configurations only.

package.json
```diff
@@ -65,6 +69,7 @@
     ]
   },
   "dependencies": {
-    "cross-fetch": "^3.1.5"
+    "cross-fetch": "^3.1.5",
+    "dotenv": "^16.3.1"
```
yen-tt, line 73: I believe this should be under devDep. I see that you only use this for the test file?

src/infra/ChatAnalyticsReporter.ts
yen-tt, line 12: nit: we should mark this as public since it's being returned in our exported `provideChatAnalytics` function

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * A class to report chat analytics data. Uses the provided API key, environment,
+ * and partition to determine the endpoint for reporting events.
+ */
+export default class ChatAnalyticsReporter {
+  public readonly apiKey: string;
```
yen-tt, line 13: nit: can we add docs for these fields since they are public?

src/models/chat/index.ts
```diff
@@ -0,0 +1,115 @@
...
+ * @public
+ */
+export type Action = (typeof ACTIONS)[number];
+
+/**
+ * The payload accepted by the Analytics Events API.
+ *
+ * @public
+ */
+export interface EventPayload {
```
yen-tt, line 70: nit: since this is public interface, we should probably add some docs for these fields

src/models/config/ChatAnalyticsConfig.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * The main configuration options for Chat Analytics.
+ *
+ * @public
+ */
+export default interface ChatAnalyticsConfig {
+  /** The API key for accessing the Analytics Events API. */
+  apiKey: string; // TODO: Add optional "token" for JWT auth
```
yen-tt, line 8: should we make an item for this TODO so we don't forget?

tests/infra/ChatAnalyticsReporter.ts
```diff
@@ -0,0 +1,63 @@
...
+  expect(response.id).toBeDefined();
+  expect(response.id).not.toBeNull();
+});
+
+it('should error when called with a bogus API key', async () => {
+  const bogusRepoter = new ChatAnalyticsReporter({
+    apiKey: 'bogus',
+    env: 'PROD',
+  });
+  await expect(bogusRepoter.report({ action: 'DIRECTIONS' })).rejects.toThrow();
```
yen-tt, line 62: nit: can we check for the error message as well?

yen-tt (commented):
also got a couple ts doc formatting errors flagged by the github action tests:

Screenshot 2023-06-22 at 10 39 42 AM

package.json
mdavish: Will do.

package.json
mdavish: Yes that's right, I'll move this to dev.

src/infra/ChatAnalyticsReporter.ts
mdavish: Yep good call.

src/models/chat/index.ts
mdavish: Good call, I totally forgot.

src/models/config/ChatAnalyticsConfig.ts
mdavish: Sure I can open up a GH issue after this PR is merged.

src/infra/ChatAnalyticsReporter.ts
mdavish: Also changed this to "region"

package.json
mdavish: @yen-tt Are you referring just to dotenv or also crossfetch?

yen-tt: dotenv only. I see that cross-fetch is used in the HttpRequester class so it's in the right dep type

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -0,0 +1,74 @@
...
+   * Report an event to the chat analytics API.
+   * @param event The event to report.
+   * @returns A promise that resolves to the response from the API.
+   * @example
+   * reporter.report({
+   *  action: 'DIRECTIONS',
+   *   chat: {
+   *   botId: 'davish-playground',
+   * },
+   * });
```
wesleyhillyext, line 51: Is there currently some provision to specify certain properties as always-passed by default?

It can wait for the final generic provideAnalytics API I guess. I was thinking about something which allows a form of currying, so the ChatAnalyticsReporter can be wrapped in reporter objects which add layers of defaults as code becomes more specific to a single site, a dingle experience, a single bot, etc.

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -0,0 +1,74 @@
...
+   * // Response
+   * {
+   *  id: '12345',
+   * }
+  */
+  public async report(event: EventPayload): Promise<EventsAPIResponse> {
+    const res = await fetch(this.endpoint, {
```
wesleyhillyext, line 58: When using the Fetch API for analytics events, passing keepalive=true seems to be the recommended approach to make sure that the event isn't cancelled by closing the page. https://developer.mozilla.org/en-US/docs/Web/API/fetch#keepalive

But, not all browsers support keepalive yet (mainly Firefox seems to be the straggler).

There's an older API called the Beacon API which is also provided for analytics events; we made sure the new Analytics Events API was compatible with Beacon (you just pass Authorization in the body instead of as a header).  https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon

src/infra/ChatAnalyticsReporter.ts
AjayBenno: ID is always returned as regardless if an event is valid or invalid we always record it. Errors will only be returned if the status is non 2xx.

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -0,0 +1,74 @@
...
+    Record<NonNullable<ChatAnalyticsConfig['env']>, string>
+  > = {
+    US: {
+      PROD: 'https://www.us.yextevents.com/accounts/me/events',
+      SANDBOX: 'https://www.sbx.us.yextevents.com/accounts/me/events',
+      QA: 'https://www.qa.us.yextevents.com/accounts/me/events',
+    },
+    EU: {
+      PROD: 'https://www.eu.yextevents.com/accounts/me/events',
+      SANDBOX: 'https://www.sbx.eu.yextevents.com/accounts/me/events',
```
AjayBenno, line 29: there is no SBX eu fyi.

src/models/chat/index.ts
AjayBenno: We had explicitly not accepted pages domain data yet since M1 was just for chat. If we need to have pages be accepted we'll need to update our endpoint to be able to start accepting data for it.

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -0,0 +1,74 @@
...
+      QA: 'https://www.qa.eu.yextevents.com/accounts/me/events',
+    }
+  };
```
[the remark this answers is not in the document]
wesleyhillyext: Oh that's odd. The new API conventions doc doesn't indicate that there should be a www component anywhere: https://docs.google.com/document/d/1pu4IhzubaChPPMX5Vmu3LeXjOfyIR6cWBx2YfNizi_E/edit

Which makes sense, as www usually indicates a website.

Can you check with Aether on this?

src/infra/ChatAnalyticsReporter.ts
mdavish: Oh I see. Should we throw an error if someone tries to do that?

mdavish: Will there ever be one?

src/infra/ChatAnalyticsReporter.ts
mdavish: Hm. Which do you think we should use?

src/infra/ChatAnalyticsReporter.ts
mdavish: Yeah this is BY FAR the most frustrating part of working with this SDK. One approach would be to add a `defaultPayload` on initialization, and then we would deep merge that payload with any payload provided directly to `.report()`.

What does everyone think of that? @AjayBenno

src/models/chat/index.ts
mdavish: [Made this into a string literal per our conversation in Slack.](https://yext.slack.com/archives/C04EW3KMLEL/p1687800787175749)

src/infra/ChatAnalyticsReporter.ts
mdavish: I think we should address this in a later follow up so we can get this out the door.

src/infra/ChatAnalyticsReporter.ts
wesleyhillyext: Pretty sure there will never be one. We're slowly moving towards getting rid of SBX even in the US partition.

yen-tt: removed

src/infra/ChatAnalyticsReporter.ts
yen-tt: updated interface and error message to account for `errors` field

src/infra/ChatAnalyticsReporter.ts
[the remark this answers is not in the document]
yen-tt: thread on this: https://yext.slack.com/archives/C73QYE16W/p1687805080039069
Once changes are made, we can deploy a patch version for `@yext/analytics` with the updated endpoints

yen-tt:
verified through test-site by sending a request to `https://www.us.yextevents.com/accounts/me/events` with proper auth headers and payload. See that the returned response have status 202. 
Screenshot 2023-06-27 at 10 38 48 AM
Screenshot 2023-06-27 at 10 38 55 AM

src/models/config/ChatAnalyticsConfig.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * The main configuration options for Chat Analytics.
+ *
+ * @public
+ */
+export interface ChatAnalyticsConfig {
+  /** The API key for accessing the Analytics Events API. */
+  apiKey: string; // TODO: Add optional "token" for JWT auth
+  /** The environment to send the requests to. */
+  env?: 'PROD' | 'SANDBOX';
```
cea2aj, line 10: Can we use "PRODUCTION" instead to be consistent with the search config?

cea2aj: I suppose the search `experienceVersion` and the Yext `env` are different concepts, but it doesn't hurt to be consistent with the vocab

cea2aj (approved):
We may want an approval from Fusion as owners of this code, but LGTM!

src/models/EventPayload.ts
```diff
@@ -0,0 +1,79 @@
...
+  /** The URL of the page where the event occurred */
+  pageUrl?: string;
+  /** The URL of the page the event is directing the visitor to. */
+  destinationUrl?: string;
+  /** A label assigned to the event, e.g. a CTA label. */
+  label?: string;
+  /** The locale of the user who generated the event. */
+  locale?: string;
+  /** The timestamp at which the event occurred, in ISO format. */
+  timestamp?: Date | string;
```
AjayBenno, line 22: Curious if we tested this.

yen-tt: when doing `JSON.stringify` on a Date object, it looks to default to ISO format already.

Screenshot 2023-06-27 at 11 35 11 AM

Screenshot 2023-06-27 at 11 35 19 AM

AjayBenno (approved):
LGTM

PR #42, opened 2023-06-29
generates sessionId when session tracking is enabled

yen-tt:
as discussed in meeting with product and fusion team, `@yext/analytics` will be in charge of providing the logic to generate `sessionId` in ULID format.

this PR updates ChatAnalyticsReporter to automatically generates and apply the sessionId to the payload iff user enable session tracking.

J=CLIP-305
TEST=manual&auto

see in test-site that sessionId is not added to payload by default
set `sessionTrackingEnabled` to true and see that sessionId is now added to the payload

see that added jest tests passed


swong1213:
lgtm

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -14,12 +15,15 @@ export class ChatAnalyticsReporter {
   private readonly apiKey: string;
   /** The endpoint to report events to, inferred by the env and region */
   private readonly endpoint: string;
+  /** Whether to enable session tracking for analytics events */
+  private readonly sessionTrackingEnabled: boolean;
 
   constructor(
-    { apiKey, env, region }: ChatAnalyticsConfig,
+    { apiKey, env, region, sessionTrackingEnabled = false }: ChatAnalyticsConfig,
```
mdavish, line 22: @yen-tt  Will the default behavior be `true` when invoked inside of other libraries like chat-ui-react?

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -47,9 +51,16 @@ export class ChatAnalyticsReporter {
...
+
+    const sessionId: string | undefined = this.sessionTrackingEnabled
+      ? (event.sessionId ?? acquireSessionId() ?? undefined)
+      : undefined;
     const res = await this.httpRequesterService.post(
       this.endpoint,
-      event,
+      {
+        ...event,
+        sessionId
```
mdavish, line 62: Should the session ID come before the event, so that you could override it in an individual call if you wanted? Or am I misremembering how spread syntax works?

src/utils/acquireSessionId.ts
```diff
@@ -0,0 +1,25 @@
+import { ulid } from 'ulid';
+
+export const SESSION_ID_KEY = 'yext_analytics_sessionId';
+
+/**
+ * Retrieves session id from session storage, or generates a new ULID to use as session id.
+ * The new id is then stored in session storage. Returns null if the sessionStorage API is
+ * unavailable (e.g. The function is running on the server for SSR).
+ */
+export function acquireSessionId(): string | null {
```
mdavish, line 10: I sort of prefer `getOrSetupSessionId` since it's crystal clear what it's doing, whereas `acquireSessionId` feels a bit more opaque. What do others think? @wesleyhillyext @nbramblett @AjayBenno @benmcginnis

src/utils/acquireSessionId.ts
```diff
@@ -0,0 +1,25 @@
+import { ulid } from 'ulid';
+
+export const SESSION_ID_KEY = 'yext_analytics_sessionId';
```
mdavish, line 3: This feels like a weird combination of camel case and snake case. Why not all camel case?

mdavish (commented):
Could we also add a blurb to the README about this, since it's pretty important and not really documented elsewhere?

src/infra/ChatAnalyticsReporter.ts
yen-tt: was going to ask you about that once this is done! I can have default behavior in Chat SDK set to true if that's what we want. My general assumption is that this should be something user opt-in instead of other way around, so I have analytics repo set it to false by default.

src/infra/ChatAnalyticsReporter.ts
yen-tt: yup, sessionId that is passed by user is handled here: https://github.com/yext/analytics/pull/42/files/a366252031511ca8b62bcaae5bdddfa583dad4a4#diff-cdc1ccc07d263bfd62a8e63ea5692787b1220210e183901c69e8fb84ba7b0fdaR55-R57
provided sessionTrackingEnabled is set to true.

src/utils/acquireSessionId.ts
yen-tt: ah, bad case of switching and copying this around.. I think chat-headless has its key for state in session storage as `yext_chat_conversation_state` so I will just follow that format and update this to `yext_analytics_session_id`

src/utils/acquireSessionId.ts
yen-tt: That's fair. I copied this name from `search-ui-react` test-site. but our old search SDK have `getOrSetupSessionId`. I can update the name

src/infra/ChatAnalyticsReporter.ts
mdavish: Perfect 👍

src/infra/ChatAnalyticsReporter.ts
mdavish: I think that we should probably make it opt out for Chat specifically just because, if you don't set it, you lose out completely on metrics like Conversation Rate. We did the same thing for the theme IIRC. Session tracking was default on, but could be toggled off.

src/utils/acquireSessionId.ts
mdavish: Yeah I'm curious what others think about the name as well. I'm open to other options!

src/infra/ChatAnalyticsReporter.ts
yen-tt: I see, then I can default to true in this repo. This config is for Chat specifically anyway

yen-tt: updated! PR now have sessionTrackingEnabled default to true

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -17,12 +18,15 @@ export class ChatAnalyticsReporter {
    * when the endpoint field from config is not provided.
    */
   private readonly endpoint: string;
+  /** Whether to enable session tracking for analytics events */
+  private readonly sessionTrackingEnabled: boolean;
 
   constructor(
-    { apiKey, env, region, endpoint }: ChatAnalyticsConfig,
+    { apiKey, env, region, endpoint, sessionTrackingEnabled = true }: ChatAnalyticsConfig,
```
benmcginnis, line 25: might want to set it to true by default in the US and false in EU

src/infra/ChatAnalyticsReporter.ts
mdavish: Great!

yen-tt:
> Could we also add a blurb to the README about this, since it's pretty important and not really documented elsewhere?

updated!

src/infra/ChatAnalyticsReporter.ts
yen-tt: done!

nbramblett (approved):

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -47,9 +51,16 @@ export class ChatAnalyticsReporter {
       Authorization: `KEY ${this.apiKey}`,
       'Content-Type': 'application/json',
     };
+
+    const sessionId: string | undefined = this.sessionTrackingEnabled
+      ? (event.sessionId ?? acquireSessionId() ?? undefined)
```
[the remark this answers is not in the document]
yen-tt: getOrSetupSessionId, specifically, window.sesionStorage.getItem return null when it doesn't fine anything. `sessionId` in payload interface is either a string or undefined. So this just converts it to undefined to be type compatible.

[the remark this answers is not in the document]
nbramblett: gotcha ok

src/infra/ChatAnalyticsReporter.ts
```diff
@@ -17,12 +18,19 @@ export class ChatAnalyticsReporter {
...
+  /** Whether to enable session tracking for analytics events */
+  private readonly sessionTrackingEnabled: boolean;
 
   constructor(
-    { apiKey, env, region, endpoint }: ChatAnalyticsConfig,
+    { apiKey, env, region = 'US', endpoint, sessionTrackingEnabled }: ChatAnalyticsConfig,
     private httpRequesterService: HttpRequesterService
   ) {
     this.apiKey = apiKey;
+    if (sessionTrackingEnabled === undefined) {
```
benmcginnis, line 29: do you want to handle null here?

yen-tt: from typescript perspective, it should only be `boolean | undefined` unless someone does something weird or or use js only. I can add the check for it and have it set to default, no harm.

benmcginnis:
```
function foo (region = 'US', sessionTrackingEnabled = region === 'US' ? true : false) {
   console.log(region, sessionTrackingEnabled);
}
foo()
// outputs: US true 
foo('EU')
// outputs: EU false
```

benmcginnis (approved):
only a minor comment

src/infra/ChatAnalyticsReporter.ts
yen-tt: updated

PR #43, opened 2023-07-05
add internalUser field to EventPayload

yen-tt:
support internalUser following https://wesleyhillyext.github.io/event/#internalUser

J=CLIP-328
TEST=manual

add `internalUser: false` to payload and see that it's included in the request


AjayBenno (approved):
LGTM

PR #45, opened 2023-07-26
support referrerUrl and visitor field in events API payload

yen-tt:
support referrerUrl and visitor field in events API payload based on doc: https://wesleyhillyext.github.io/event/#visitor

J=CLIP-374
TEST=manual

verified the report request worked with referrerUrl and visitor in the payload:
Screenshot 2023-07-28 at 2 31 36 PM
Screenshot 2023-07-28 at 2 32 12 PM


src/models/EventPayload.ts
```diff
@@ -14,6 +14,8 @@ export interface EventPayload {
   pageUrl?: string;
   /** The URL of the page the event is directing the visitor to. */
   destinationUrl?: string;
+  /** The URL of the page which the visitor came from prior to the event. */
```
AjayBenno, line 17: For both this and pageUrl can we add that the format must be a URI?

src/models/EventPayload.ts
```diff
@@ -78,4 +80,13 @@ export interface EventPayload {
     /** The algorithm to use to anonymize the IP address after collection. */
     algorithm?: string;
   };
+  /**
+   * Information used to associate analytics with a particular user.
+   *
+   * @remarks
+   * Keys are visitor methods and values are visitor IDs
```
AjayBenno, line 87: We could also add the regex for a visitor method `^[a-zA-Z0-9_-]{1,16}$ ` and visitor ID `^[\x20-\x7E]{1,64}$`

AjayBenno (commented):
Some doc nits

src/models/EventPayload.ts
```diff
@@ -10,10 +10,12 @@ export interface EventPayload {
   action: Action;
   /** Unique identifier to tie together events in a single browsing session */
   sessionId?: string;
-  /** The URL of the page where the event occurred */
+  /** The URI of the page where the event occurred */
```
AjayBenno, line 13: Hmn, i think it's not quite right to call it the URI of the page. All URL's are of URI format, so maybe we can keep the word URL and mention that it must be in URI format? 

Thinking about this again it's probably sufficient to just say URL. If you try passing in a non-url you'll get a 4xx error from the events API mentioning the format is wrong so seems intuitive. 

Sorry let's revert this back to URL across the board. LGTM otherwise.

AjayBenno (approved):
LGTM

PR #48, opened 2023-08-21
EventsSDK: Set up dev/unified-sdk

ejaffee01:
Delete unused code and reorganize file structure for new version of the Analytics Events SDK.

J=FUS-5866
R=mtian, abenno
TEST=none

Thought to npm run build, but as index.ts is not set up this will fail.


src/index.ts
```diff
@@ -1,95 +1 @@
...
-  AnalyticsService,
-  PagesAnalyticsService,
-  SearchAnalyticsService,
-  ConversionTrackingService,
-  ChatAnalyticsService
-} from './services';
-
-export * from './utils';
\ No newline at end of file
+/** To be implemented in https://yexttest.atlassian.net/browse/FUS-5874 */
```
mtian725, line 1: Let's do 

`export * from '.'`

just so it builds.

mtian725 (changes requested):
Let's also change the version to "1.0.0" in the package.json

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/analytics",
-  "version": "0.6.4",
+  "version": "1.0.0",
   "description": "An analytics library for Yext",
   "author": "slapshot@yext.com",
```
AjayBenno, line 5: fusion@yextcom

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/analytics",
-  "version": "0.6.4",
+  "version": "1.0.0",
```
AjayBenno, line 3: 1.0.0-beta.0

package.json
AjayBenno, line 1: can we add the --passWithNoTests argument so that we can have the tests pass on this commit? Should be able to add a new line to the jest object in this file like 
`"passWithNoTest": true,`

AjayBenno (approved):
LG

PR #49, opened 2023-08-21
EventsSDK: create AnalyticsConfig interface

swong1213:
Create AnalyticsConfig interface.

J=FUS-5867
TEST=auto

### DIFF
--- a/src/AnalyticsConfig.ts
+++ b/src/AnalyticsConfig.ts
@@ -7,7 +7,13 @@ import {Region} from './models';
  * @public
  */
 export interface AnalyticsConfig {
+    /** The API Key or OAuth for accessing the Analytics Events API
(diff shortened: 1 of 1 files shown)


src/AnalyticsConfig.ts
```diff
@@ -0,0 +1,23 @@
+import {Environment} from './models';
+import {Region} from './models';
+
+/**
+ * The main configuration options for Analytics Events.
+ *
+ * @public
+ */
+export interface AnalyticsConfig {
+    key?: string;
```
mtian725, line 10: Lets add a comment explaining that only one or other needs to be set

swong1213: Added comment

mtian725 (approved):

PR #50, opened 2023-08-21
EventsSDK: create AnalyticsEventService

swong1213:
Create AnalyticsEventService interface that will be implemented by AnalyticsEventReporter

J=FUS-5868
TEST=auto


src/AnalyticsEventService.ts
```diff
@@ -0,0 +1,23 @@
+import {AnalyticsConfig} from "./AnalyticsConfig";
+import {EventAPIResponse, EventPayload} from "./models";
```
mtian725, line 2: I think you might need to rebase.

swong1213: Hmm rebasing adds all my previous commits to the list but I think I have to in order for the checks to pass

src/AnalyticsEventService.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * Reports an analytics event. Operand will throw an error if the request
+     * fails. The request will be sent via Beacon API so that requests are
+     * non-blocking.
+     *
+     * @param payload - desired values to be applied. The new payload will
+     * override any overlapping values.
+     */
+    report(payload: object): Promise<EventAPIResponse>;
```
alinayh, line 22: I might misremembered this from sprint planning: Do we want EventAPIResponse with or without Promise? @mtian725

src/AnalyticsEventService.ts
swong1213: I think I'm just gonna start over with a new PR...

src/AnalyticsEventService.ts
mtian725: @alinayh you brought up a good point that I completely overlooked. Beacon doesnt return a promise. It just returns a boolean value but that boolean is whether or not the user agent is added to the request information successfully. It will still throw an error if the request fails, but we have no way to get the response body. 

I discussed this with Ajay last night and we discuss how we might want to go about this because fetch with the keepalive flag is functionally equivalent, but firefox does not support this feature.

mtian725 (approved):

PR #51, opened 2023-08-22
EventsApi: Implement merge

ejaffee01:
Implement merge function used for combining request json into an existing EventPayload object. Merges using the following conventions:

- Merging an existing key with null/undefined deletes the key
- Merging an existing key with non-null value update the value
- Merging a non-existing key adds the key/value

The above is applied recursively to nested objects. This change also updates the bot type in EventPayload. 

J=FUS-5869
R=abenno, mtian
TEST=unit

Wrote three unit tests and all pass.

### DIFF
--- a/src/merge.ts
+++ b/src/merge.ts
@@ -14,7 +14,7 @@ export const merge = (
   original: EventPayload,
   newValues: Partial<Record<keyof EventPayload, unknown>>
 ): EventPayload => {
-  const copy: Record<string, unknown> = { ...original };
+  const copy = { ...original };
   const stack: Array<[Record<string, unknown>,
      Partial<Record<keyof EventPayload, unknown>>]> = [[copy, newValues]];
 
--- a/src/merge.ts
+++ b/src/merge.ts
@@ -1,0 +1,25 @@
+import { EventPayload } from './EventPayload';
+
+/** Used for merging request JSON into an existing EventPayload. Returns a new EventPayload object */
+export const merge = (json1: EventPayload, json2: Record<string, unknown>): EventPayload => {
+    const copy: Record<string, unknown> = { ...json1 };
+
+    for (const key in json2) {
+        if (json2.hasOwnProperty(key)) {
+            const value = json2[key];
+            if (value === null || value === undefined) {
+                delete copy[key];
+            } else if (typeof value !== 'object') {
+                copy[key] = value;
+            } else {
+                copy[key] = merge(copy[key] as EventPayload, value as Record<string, unknown>);
+            }
+        }
+    }
+
+    const result: EventPayload = {
+        ...copy,
+        action: copy.action as EventPayload['action']
+    };
+    return result;
+};
--- a/src/merge.ts
+++ b/src/merge.ts
@@ -10,28 +10,28 @@ import { EventPayload } from './EventPayload';
  *  Merging a non-existing key adds the key/value.
  *  When the value is an object, the above is applied iteratively.
  */
-export const merge = (
-  original: EventPayload,
-  newValues: Partial<Record<keyof EventPayload, unknown>>
-): EventPayload => {
+export const merge = (original: EventPayload, newValues: Partial<PartialPayload>): EventPayload => {
   const copy = { ...original };
-  const stack: Array<[Record<string, unknown>,
-     Partial<Record<keyof EventPayload, unknown>>]> = [[copy, newValues]];
+  const stack: Array<[PartialPayload, PartialPayload]> = [[copy, newValues]];
 
   while (stack.length) {
     const [target, source] = stack.pop() || [];
     if (target && source) {
       Object.keys(source).forEach((key) => {
-        const value = source[key as keyof EventPayload];
+        console.log('Stack: ' + JSON.stringify(stack));
+        const EventKey = key as keyof EventPayload;
+        console.log('key' + EventKey);
+        const value = source[EventKey];
         if (value === null || value === undefined) {
-          delete target[key];
+          delete target[EventKey];
         } else if (typeof value !== 'object') {
-          target[key] = value;
+          target[EventKey] = value;
         } else {
-          if (!target[key]) {
-            target[key] = {};
+          if (!target[EventKey]) {
+            target[EventKey] = {};
           }
-          stack.push([target[key] as Record<keyof EventPayload, unknown>, value]);
+          console.log('target: ' + JSON.stringify(target) + '\nsource: ' + JSON.stringify(source));
+          stack.push([target[EventKey] ?? {}, value]);
         }
       });
     }
@@ -42,3 +42,5 @@ export const merge = (
     action: copy.action as EventPayload['action']
   };
 };
+
+type PartialPayload = Partial<Record<keyof EventPayload, unknown>>;
--- a/src/merge.ts
+++ b/src/merge.ts
@@ -1,14 +1,21 @@
 import { EventPayload } from './EventPayload';
 
-/** Used for merging request JSON into an existing EventPayload.
+/**
+ * Used for merging request JSON into an existing EventPayload.
  * Returns a new EventPayload, with the original object uneffected.
+ * The merge occurs using the following conventions:
+ *
+ *  Merging an existing key with null/undefined deletes the key.
+ *  Merging an existing key with non-null value update the value.
+ *  Merging a non-existing key adds the key/value.
+ *  When the value is an object, the above is applied recursivley.
  */
-export const merge = (original: EventPayload, newValues: Record<string, unknown>): EventPayload => {
+export const merge = (original: EventPayload, newValues: Partial<Record<keyof EventPayload, unknown>>): EventPayload => {
   const copy: Record<string, unknown> = { ...original };
 
   for (const key in newValues) {
     if (newValues.hasOwnProperty(key)) {
-      const value = newValues[key];
+      const value = (newValues as Record<string, unknown>)[key];
       if (value === null || value === undefined) {
         delete copy[key];
       } else if (typeof value !== 'object') {
--- a/src/merge.ts
+++ b/src/merge.ts
@@ -18,9 +18,7 @@ export const merge = (original: EventPayload, newValues: Partial<PartialPayload>
     const [target, source] = stack.pop() || [];
     if (target && source) {
       Object.keys(source).forEach((key) => {
-        console.log('Stack: ' + JSON.stringify(stack));
         const EventKey = key as keyof EventPayload;
-        console.log('key' + EventKey);
         const value = source[EventKey];
         if (value === null || value === undefined) {
           delete target[EventKey];
@@ -30,7 +28,6 @@ export const merge = (original: EventPayload, newValues: Partial<PartialPayload>
           if (!target[EventKey]) {
             target[EventKey] = {};
           }
-          console.log('target: ' + JSON.stringify(target) + '\nsource: ' + JSON.stringify(source));
           stack.push([target[EventKey] ?? {}, value]);
         }
       });
--- a/src/merge.ts
+++ b/src/merge.ts
@@ -11,24 +11,24 @@ import { EventPayload } from './EventPayload';
  *  When the value is an object, the above is applied iteratively.
  */
 export const merge = (original: EventPayload, newValues: PartialPayload): EventPayload => {
-  const copy = { ...original };
+  const copy = JSON.parse(JSON.stringify(original));
   const stack: Array<[PartialPayload, PartialPayload]> = [[copy, newValues]];
 
   while (stack.length) {
     const [target, source] = stack.pop() ?? [];
     if (target && source) {
       Object.keys(source).forEach((key) => {
-        const EventKey = key as keyof EventPayload;
-        const value = source[EventKey];
+        const eventKey = key as keyof EventPayload;
+        const value = source[eventKey];
         if (value === null || value === undefined) {
-          delete target[EventKey];
+          target[eventKey] = undefined;
         } else if (typeof value !== 'object') {
-          target[EventKey] = value;
+          target[eventKey] = value;
         } else {
-          if (!target[EventKey]) {
-            target[EventKey] = {};
+          if (!target[eventKey]) {
+            target[eventKey] = {};
           }
-          stack.push([target[EventKey] as PartialPayload, value]);
+          stack.push([target[eventKey] as PartialPayload, value]);
         }
       });
     }
--- a/src/merge.ts
+++ b/src/merge.ts
@@ -28,7 +28,7 @@ export const merge = (original: EventPayload, newValues: Partial<PartialPayload>
           if (!target[EventKey]) {
             target[EventKey] = {};
           }
-          stack.push([target[EventKey] as PartialPayload ?? {}, value]);
+          stack.push([target[EventKey] as PartialPayload, value]);
         }
       });
     }
--- a/src/merge.ts
+++ b/src/merge.ts
@@ -28,7 +28,7 @@ export const merge = (original: EventPayload, newValues: Partial<PartialPayload>
           if (!target[EventKey]) {
             target[EventKey] = {};
           }
-          stack.push([target[EventKey] ?? {}, value]);
+          stack.push([target[EventKey] as PartialPayload ?? {}, value]);
         }
       });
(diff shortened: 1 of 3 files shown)


src/merge.ts
```diff
@@ -0,0 +1,25 @@
+import { EventPayload } from './EventPayload';
+
+/** Used for merging request JSON into an existing EventPayload. Returns a new EventPayload object */
+export const merge = (json1: EventPayload, json2: Record<string, unknown>): EventPayload => {
```
mtian725, line 4: I originally used json1 and json2 as a placeholder. We should use more descriptive naming here. Maybe original and newValues?

src/merge.ts
```diff
@@ -0,0 +1,25 @@
+import { EventPayload } from './EventPayload';
+
+/** Used for merging request JSON into an existing EventPayload. Returns a new EventPayload object */
```
mtian725, line 3: Add to this that the original object will not be effected.

src/merge.ts
```diff
@@ -0,0 +1,25 @@
...
+                copy[key] = value;
+            } else {
+                copy[key] = merge(copy[key] as EventPayload, value as Record<string, unknown>);
+            }
+        }
+    }
+
+    const result: EventPayload = {
+        ...copy,
+        action: copy.action as EventPayload['action']
```
mtian725, line 22: Why do you need to do this? Does `const result: EventPayload = copy` not work?

tests/merge.test.ts
```diff
@@ -0,0 +1,100 @@
+import { EventPayload } from '../src/EventPayload';
+import { merge } from '../src/merge';
+
+describe('Merge Function', () => {
+    it('should merge example 1 correctly', () => {
```
mtian725, line 5: We shouldn't have the description reference the spec. Let's rename them to something like:
- "merging primitive values"
- "merging objects (no overlap)"
- "merging objects (with overlap)"
If you think of a better descriptor, go for it. These are just possible ideas.

src/merge.ts
ejaffee01: Unfortunately it does not. Doing that or 
 `const result: EventPayload = {...copy};`
results in 
`Property 'action' is missing in type '{ [x: string]: unknown; }' but required in type 'EventPayload'.ts(2741)`

mtian725: Gotcha. Also, isn't the spread operator `...` result in this being a shallow copy?

tests/merge.test.ts
```diff
@@ -0,0 +1,118 @@
...
+        botId: 'botId',
+        responseId: 'responseId',
+      },
+    };
+
+    expect(result).toEqual(expected);
+    expect(event1).toEqual(uneffectedEvent1); // confirm event1 uneffected
+  });
+
+  it('should merge primitives and objects (with overlap) correctly', () => {
```
mtian725, line 71: We should also add a test to cover merging with values not in the EventPayload.

src/merge.ts
```diff
@@ -0,0 +1,27 @@
+import { EventPayload } from './EventPayload';
+
+/** Used for merging request JSON into an existing EventPayload.
```
AjayBenno, line 3: Can you list out the three cases for the merge?

src/merge.ts
ejaffee01: It makes a deep copy at the top level, but if it has nested objects then the children are shallow copies: https://www.educative.io/answers/what-is-the-spread-operator-in-javascript

However this is not an issue because we do not ever change the original.

tests/merge.test.ts
ejaffee01: Discussed offline. This validation is not really possible without having a hardcoded data structure of the possible keys and comparing that to what is in the record. This is a complicated process and not really necessary considering that the request will eventually be sent to the Events API, which will throw an error if the body is invalid (and will describe what is invalid about it). 

However, by changing the type of the record to be Partial<Record<keyof EventPayload, unknown>> we can provide some type checking in the IDE if the input is not valid for EventPayload (likely only for typescript users).

src/merge.ts
```diff
@@ -0,0 +1,44 @@
...
+ *  Merging an existing key with non-null value updates the value.
+ *  Merging a non-existing key adds the key/value.
+ *  When the value is an object, the above is applied iteratively.
+ */
+export const merge = (
+  original: EventPayload,
+  newValues: Partial<Record<keyof EventPayload, unknown>>
+): EventPayload => {
+  const copy: Record<string, unknown> = { ...original };
+  const stack: Array<[Record<string, unknown>,
```
mtian725, line 18: nit: don't need to state the type. It's inferable from the assignment.

src/merge.ts
```diff
@@ -0,0 +1,44 @@
...
+     Partial<Record<keyof EventPayload, unknown>>]> = [[copy, newValues]];
+
+  while (stack.length > 0) {
+    const [target, source] = stack.pop() || [];
+    if (target && source) {
+      Object.keys(source).forEach((key) => {
+        const value = source[key as keyof EventPayload];
+        if (value === null || value === undefined) {
+          delete target[key];
+        } else if (typeof value !== 'object' || value === null) {
```
mtian725, line 28: Isn't `value === null` here redundant with the if above?

src/merge.ts
```diff
@@ -0,0 +1,44 @@
...
+          target[key] = value;
+        } else {
+          if (!target[key]) {
```
mtian725, line 31: What's the point of this? Looking at this, it seems as if the key doesn't exist in the target, you're setting it up as an empty object. Is this correct?

src/merge.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export const merge = (
+  original: EventPayload,
+  newValues: Partial<Record<keyof EventPayload, unknown>>
+): EventPayload => {
+  const copy: Record<string, unknown> = { ...original };
+  const stack: Array<[Record<string, unknown>,
+     Partial<Record<keyof EventPayload, unknown>>]> = [[copy, newValues]];
+
+  while (stack.length > 0) {
```
mtian725, line 21: Up to you for code readability, but 0 is falsy in Javascript/Typescript

src/merge.ts
```diff
@@ -0,0 +1,44 @@
...
+    const [target, source] = stack.pop() || [];
```
mtian725, line 22: Do you need to do `stack.pop() || []`? There should always be a value in the stack as long as the loop is running.

src/merge.ts
ejaffee01: That is correct. It gets set up as an empty object and added to the stack with the corresponding value. On the next iteration it will add the value as necessary.

src/merge.ts
ejaffee01: While that is true, typescript still complains about it without

Type '[Record<string, unknown>, Partial<Record<keyof EventPayload, unknown>>] | undefined' is not an array type.
when I do const [target, source] = stack.pop();

I also tried const [target, source] = stack.pop()!;
but that has a warning 
Forbidden non-null assertion.eslint[@typescript-eslint/no-non-null-assertion](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/no-non-null-assertion.md)

src/merge.ts
ejaffee01: Yup

src/merge.ts
mtian725: Small correction, if you want to keep it as and Array<tuple>, you need to do something like the below
`const stack = new Array([copy, newValues])`

src/merge.ts
mtian725: Then in that case let's use ?? instead of ||

src/merge.ts
mtian725: As discussed offline, this doesn't work as the inferred type is not exactly the same.

src/merge.ts
```diff
@@ -0,0 +1,44 @@
...
+        const value = source[key as keyof EventPayload];
+        if (value === null || value === undefined) {
+          delete target[key];
+        } else if (typeof value !== 'object') {
+          target[key] = value;
+        } else {
+          if (!target[key]) {
+            target[key] = {};
+          }
+          stack.push([target[key] as Record<keyof EventPayload, unknown>, value]);
```
mtian725, line 34: Is this supposed to be `Record<string, unknown>` rather than `Record<keyof EventPayload, unknown> `

src/merge.ts
ejaffee01: Going with `const stack: Array<[PartialPayload, PartialPayload]> = [[copy, newValues]];` as discussed offline

src/merge.ts
ejaffee01: simplified as discussed offline.

src/merge.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export const merge = (original: EventPayload, newValues: PartialPayload): EventPayload => {
+  const copy = { ...original };
+  const stack: Array<[PartialPayload, PartialPayload]> = [[copy, newValues]];
+
+  while (stack.length) {
+    const [target, source] = stack.pop() ?? [];
+    if (target && source) {
+      Object.keys(source).forEach((key) => {
+        const EventKey = key as keyof EventPayload;
```
nbramblett, line 21: seems a bit odd to capitalize EventKey when it's a variable and not a type or constructor

src/merge.ts
ejaffee01: Appears I was wrong about this: [See example](https://www.typescriptlang.org/play?#code/FAYw9gdgzgLgBAWwKYCcDmS4F44AowoCWahEAhgDYBccAogG5IQwAKZAnhWGQCYA0cCEgDuANUoBXJFBpsUMQpTaduPAJQ0GTVhy69sAPjgBvYHDjho8cAAd22E3AB0LgsVKU4AXwDcZi5CwcLBkIADWNACCKCgcADwA2nIKSrqqAsmKFMp6PAC6RjgJCbbsAkJiktJ5eX7+wgAWhBSYuCHhTi0QaDANaib+5pZBCTBk6EgwAlBgEiggSHkO7WFONmA2uP0A-NtwCbWDcIQAZnhjE-AAZFfBs-NI-abmL3AA8gBGAFZIIDBOYSQ7CgbXuCzUThOBFooQauFwgPY-SwRmerxew3gWmYAGkgQ5EXAyFA4IiwGdsToVLw-OiMYF4PQqsswUgEpS8exDnTzKc8EyKFJsFgcBAJBQKHAAD5SuACoUinASCA8JAnUhIdQDHkvVUtGCYC4YGDsxi4oHcnleOBIChQTB83Awdg2JDkuXMgCEIrgAHIwN9fjBfU8jnSjZNTdpOUscPKkLSrTa7Zg0TrHZ6IyaORbQzrXlmo+auQ5jL4w+ivBXXis1hIoHDRuNjUWYDGiSTMqlqfwPYLFmpE3Sq8PB0cR95gP4UJM5hBtS8XE5SnwjqEFJAaKUnOvCJAO3QzVTcglfbvIL68v5y+XQAy4G4SORJThjDANjRjEJYDQAERQsBfy8KsAHoQLgAABGAoAAWmICACCQYBkAmfAiCfSgBDfD8TG-GA-w+cYgK8NQp2GMAWk6MA0DQ9xnycd8bCcPD+jAuBf0IlBfynUgDRQE5QkwSkclUBcQIAKnEuAABUGkwetUCJP493nRpCBABoLDIBSeDgXpMCQI8BCQJw0CcOBIgAEUsgB9aS3hsgBhSIACVpIfFAZIACQAVQAWQAIQAZRsnyWHMcSQP8c8IBoWAiG6RMJKknyIEIABHIVCFVZhTkIRT3z0-K9OoyY5M8wztBJUgiWCUg0BaOAPhQMBhCgergmkdr90i-x7SgbqIAASR4bY4pgBK0CSySZLkuAfJcgAZB8zn0uAbDIDA4EaVBDTmyrmAfEAQDmGddN68wNowHyUAoMbggm+rpqk2TMAW5b3TWq69oMo9jhJHhCBnZTuj0ub6EIdr30899zIuuBVVgDwNwgG67vGybntmt6lpWsHMG+7amg0-G5Uhwhoa05A4BOFqEHWogCBK0mDv+OB4ZnE5UBnFA0fu+Knv8ZKLLgCgyA+W0O3atAhF0wq1tZ4zTPMsg4Ec6TIlF8XbThqLzDFiX0YezGhZm17RbAEBKEwT65oUzzGjAOAMCEWIDTl-aj11-wuCtlp+cexLTZeuaFGQEIEBsIl4DUkmFb+y2TpiTUBBqoagreGmCAQMg2fhsPpDGSP7ss3PMFlAWg-MYWAHU5P0mHPe0f7SZnKBxXgd0PjAeBdwh51vfMbuYADk3q5moaIChFAc5Rolu4keA1ohqGCABwz1MwMgVSalq2tQQfd9a+0UEiF2R8-I5hfN5rj8U4kZhARR3e2inNPj7RD6HveT9HwXXmvnNW++9PKMBQINKWltn6alfr0FmXt2Z6xeMAk+ohUCDT-lXF4gCbaujdh1KAwIDT0wflAsuulhBv3gZ-RBRwwBQEwVNK+Zs5obFQLnAhRCkD0zARA0hT9yGwPfk3ZgX8HxQDQeAlSjChxwBwQjDeCxIECJfpQuBH9RG0NeKqCGCwZHMJDpgHRm8LBiwGso6BFCqEaLzkg8wxiFiOTMQwjG-9sEsPkifIk58LGCLUcI36ND4bmHtmfbQ+jzDlnHuJQYUkABiTM1oAE0kAAA9rAUHyodIKlkcTVQgPQCijBdI1RsQIXokMW7bwfIGP4iAyA2BsPVWJpNyDhzxr0ehmAcl5OZsvdBKktaGxJLbLpcAelQCcLEpBIBMnaCCjwMI90XK-AIDwOIlc0DTEDmgAwWNJ6AytgaEkO0G7UMOpUtabcO5414qgZ8SkFD93YIfO5KBnw+V-jQbuFEkDb2ei0uuTBW6rJQLpduCAc5EAAF7SEQB3QgNhGqsygOUuaYoEAS08u6FFILwBgrgDOGwbcqpTJeIYm0f1LkzkEcSY4ZwKYtx4BIJF6lBEVJJDnCA9gC6TOmf4cAyoL6CAkJi1AALyXzSjoVAAjAABnWmQIGnLc4aQ6idWAYB6abNJECEkhUdXxn1U7fh0ChHnP+C0zkJJxiYCtvaOC0AmBQ0IIwHwfYpA2pnFpB19poAU1dSZflQx6zvgQNJTaLi4ArPxeszZ2zJp7ODi0ny0qnbysVcq+pMA1Wgw1WG42HVETGpFWK0BVQS2mrLuamxZKIpwGtUSb19qkCOr9S6xgdatEWFDVq8Q-Yo0xrWRsnZ5RRVYqTdE7GcAUnpJtLlZ0zNY4BMpc3fFbd1gqkmRFJB2gKbsDGkcWUaZ3EUoQIvcWLQBD5q1agGC9oYBjA+Mihd9ghqWSzo3Ayr7fGqOsSI2x1Z50KGdCNVxWDXgTnMMeis8jCCQovc+79IG30fofU1Hlc1CGwG4eZWSlTKnVLeQ899YjXh7udD5bKNAMVYtkZErG5shrhV4DwNuJJp4WqcPDRF90T3yOY-QAALESHgbGuq-pgf4rj3bzCsfY+BphACPFEgoGgNwvR6aFQUszbekB2AIEILC0mzHRPifMWQE4fEAgSiDCpMRlB1NEE0xE7wEr62T2nrPQZOldMDTIQaIkz52AKBACcqhqsNryHUuKcYcB7ZdpiZKiCM4oVhCgFavVTbMArwpkzZAnSeA2p3ka7LZNV6eXfXyyVnJFWPvuTQAABgAPQSGQGC0LIgwQAFpypggAThsjBPIxgZV8BlQANi8AAEka4luA-ahQbXq+8prrWAA6qSABMfXNsAHZaAjbGxNoTM3GuxJaRBGwEhn3qWDeVvLKBlmgrjaOwt3Qk1VmAM6V0cAuzZDSPoHA-24hDrBXEMkFIjwiV4AIZUYQELCAgAYPZQA)

Updating to make a deep copy

mtian725 (approved):

PR #53, opened 2023-08-22
EventsSDK: create post() util function

swong1213:
Create post() util function. The function will be updated in a later story (https://yexttest.atlassian.net/browse/FUS-5878) based on discussions from this thread:
https://yext.slack.com/archives/G01SSAYJVM4/p1692731460458789?thread_ts=1692715631.720839&cid=G01SSAYJVM4

This story just creates a basic post() function to unblock other stories in this epic.

J=FUS-5878
TEST=auto

### DIFF
--- a/src/post.ts
+++ b/src/post.ts
@@ -1,0 +1,10 @@
+import {EventPayload} from './EventPayload';
+
+const post = (
+    url: string,
+    body: EventPayload
+) => {
+    return navigator.sendBeacon(url, JSON.stringify(body));
+}
+
+export default post;


src/post.ts
```diff
@@ -0,0 +1,10 @@
+import {EventPayload} from './EventPayload';
+
+const post = (
```
mtian725, line 3: nit: can we add a comment explaining how this will be used? Also, maybe consider marking it as a to-do stating that this will change significantly.

src/post.ts
```diff
@@ -0,0 +1,10 @@
...
+    url: string,
+    body: EventPayload
+) => {
+    return navigator.sendBeacon(url, JSON.stringify(body));
```
mtian725, line 7: Are you not able to just pass down `body`? sendBeacon does accept object as it's data param.

swong1213: There's an error if I just pass down `body` :

`Argument of type 'EventPayload' is not assignable to parameter of type 'BodyInit | null | undefined'.`

Not sure if it's just an issue with the IDE, but most examples I'm seeing online stringifies the JSON data (e.g. https://xgwang.me/posts/you-may-not-know-beacon/ or https://www.w3.org/TR/beacon/#example-1), or creates a FormData or Blob object (https://medium.com/@laura.nickerson/javascripts-sendbeacon-de1cc1c97b69)

src/post.ts
swong1213: Done. I also included in the todo that we should also update the description.

PR #54, opened 2023-08-23
Implement AnalyticsEventReporter constructor

alinayh:
Set up the constructor for the AnalyticsEventReporter. Validate AnalyticsConfig only have one and only one of API Key or Bearer Token.

J=5871
TEST=auto

### DIFF
--- a/src/AnalyticsEventReporter.ts
+++ b/src/AnalyticsEventReporter.ts
@@ -16,15 +16,14 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
      * @param payload - (optional) desired event values to report
      */
     constructor(config: AnalyticsConfig, payload?: EventPayload) {
-        const configIsValid: boolean = ('key' in config || 'bearer' in config) 
-            && !('key' in config && 'bearer' in config);
+        const apiKeyIsSet = typeof config.key === 'string' && config.key.trim().length !== 0;
+        const bearerTokenIsSet = typeof config.bearer === 'string' && config.bearer.trim().length !== 0;
+        const configIsValid = (apiKeyIsSet || bearerTokenIsSet) && !(apiKeyIsSet && bearerTokenIsSet);
         if (!configIsValid) {
             throw new Error("Provide one and only one of API Key or Bearer Token.")
         }
         this.config = config;
-        if (payload !== undefined) {
-            this.payload = payload;
-        }
+        this.payload = payload;
     }
 
     with(payload: EventPayload): AnalyticsEventService {
--- a/src/AnalyticsEventReporter.ts
+++ b/src/AnalyticsEventReporter.ts
@@ -1,0 +1,37 @@
+import { AnalyticsEventService } from "./AnalyticsEventService"
+import { AnalyticsConfig } from "./AnalyticsConfig";
+import { EventPayload } from "./EventPayload";
+import { EventAPIResponse } from "./EventAPIResponse";
+
+/** Represents an reporter is responsible for reporting analytics events. */
+export class AnalyticsEventReporter implements AnalyticsEventService {
+
+    private config: AnalyticsConfig;
+    private payload?: EventPayload;
+
+    /**
+     * @param config - necessary analytics config: Must provide one and only
+     * one of API Key or Bearer Token.
+     * 
+     * @param payload - (optional) desired event values to report
+     */
+    constructor(config: AnalyticsConfig, payload?: EventPayload) {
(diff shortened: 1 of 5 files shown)


src/AnalyticsEventReporter.ts
```diff
@@ -0,0 +1,37 @@
...
+    private payload?: EventPayload;
+
+    /**
+     * @param config - necessary analytics config: Must provide one and only
+     * one of API Key or Bearer Token.
+     * 
+     * @param payload - (optional) desired event values to report
+     */
+    constructor(config: AnalyticsConfig, payload?: EventPayload) {
+        const configIsValid: boolean = ('key' in config || 'bearer' in config)
```
mtian725, line 19: I would advise against using `<key> in <object>` because this is true if the value is undefined. `in` just checks if the key exists, and `variable?: <some type>` is equal to `variable?: <some type> | undefined`.

tests/AnalyticsEventReporter.test.ts
```diff
@@ -0,0 +1,16 @@
+import { AnalyticsConfig } from "../src/AnalyticsConfig";
+import { AnalyticsEventReporter } from "../src/AnalyticsEventReporter";
+
+it("Invalid config with no authorization field", () => {
```
mtian725, line 4: Let's also add tests with valid configs.

src/AnalyticsEventReporter.ts
```diff
@@ -0,0 +1,37 @@
...
+     * @param payload - (optional) desired event values to report
+     */
+    constructor(config: AnalyticsConfig, payload?: EventPayload) {
+        const configIsValid: boolean = ('key' in config || 'bearer' in config) 
+            && !('key' in config && 'bearer' in config);
+        if (!configIsValid) {
+            throw new Error("Provide one and only one of API Key or Bearer Token.")
+        }
+        this.config = config;
+        if (payload !== undefined) {
```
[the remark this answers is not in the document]
alinayh: Done

src/AnalyticsEventReporter.ts
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * @param config - necessary analytics config: Must provide one and only
+     * one of API Key or Bearer Token.
+     * 
+     * @param payload - (optional) desired event values to report
+     */
+    constructor(config: AnalyticsConfig, payload?: EventPayload) {
+        const apiKeyIsSet = typeof config.key === 'string' && config.key.trim().length !== 0;
+        const bearerTokenIsSet = typeof config.bearer === 'string' && config.bearer.trim().length !== 0;
+        const configIsValid = (apiKeyIsSet || bearerTokenIsSet) && !(apiKeyIsSet && bearerTokenIsSet);
```
AjayBenno, line 21: could just be apiKeyIsSet ^ bearerTokenIsSet right? (xor operator)

alinayh: Typescript does not support ^ for boolean values. I can do apiKeyIsSet !== bearerTokenIsSet, but think its readability is bad.

mtian725: ^ is a bitwise operator. You can consider converting the boolean into a number, but that also might be a bit awkward for readability.

src/AnalyticsEventReporter.ts
```diff
@@ -0,0 +1,36 @@
...
+    private payload?: EventPayload;
+
+    /**
+     * @param config - necessary analytics config: Must provide one and only
+     * one of API Key or Bearer Token.
+     * 
+     * @param payload - (optional) desired event values to report
+     */
+    constructor(config: AnalyticsConfig, payload?: EventPayload) {
+        const apiKeyIsSet = typeof config.key === 'string' && config.key.trim().length !== 0;
```
mtian725, line 19: There is no need to check if it is a string. If you are doing this for the sake of confirming it is neither undefined or null, you can just use `??`, which functions exactly the same as coalesce in mySQL.

I would also not bother with trimming the whitespace and validating if they are putting in a string of spaces. If the user does that, that's on them.

alinayh: Actually, how about config.key !== undefined? It considers cases when key is not set in config and when key is specified as undefined.

mtian725: That seems fine to me.

alinayh (commented):
I'll close this out and reopen another one. Not sure how to deal with this

PR #57, opened 2023-08-23
EventsSDK: Update post() to use fetch + keepAlive

swong1213:
Update the post() util function to use fetch and keepAlive by default, unless the forceFetch flag is false and the browser does not support keepAlive (it should be supported by all browsers except Firefox and Firefox for Android).

J=FUS-5878
TEST=auto

Wrote jest tests and ran npm run test.


tests/post.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    const url = 'https://dev.us.yextevents.com/accounts/me/events';
+
+    const optionsA = {method: 'POST', body: JSON.stringify(eventPayloadA), keepalive: true};
+
+    fetchMock.post(url, JSON.stringify(eventResponseA));
+
+    it('should send fetch request', async () => {
+        // forceFetch: true
+        await post(url, eventPayloadA, true);
+        expect(fetchMock.calls('https://dev.us.yextevents.com/accounts/me/events')[0][1]).toEqual(optionsA);
```
swong1213, line 39: I'm having trouble with mocking the response of the fetch call, so I'm only checking that fetch gets called correctly right now. Putting up a PR to be reviewed first while I look into it.

src/post.ts
```diff
@@ -1,20 +1,27 @@
...
+const post = async (
     url: string,
-    body: EventPayload
+    body: EventPayload,
+    forceFetch: boolean,
 ) => {
-    return navigator.sendBeacon(url, JSON.stringify(body));
+    const agent = body.browserAgent;
+    // fetch + keepalive is not supported in Firefox
+    let useBeacon = (agent && agent.browser && agent.browser.startsWith('Firefox')) || (navigator.userAgent && navigator.userAgent.includes('Firefox'));
```
AjayBenno, line 20: Hmn does the user setting agent in the event payload really tell us if this is on firefox?

src/post.ts
```diff
@@ -1,20 +1,27 @@
...
+ * Used by the AnalyticsEventReport report() method to send an analytics event request via
+ * fetch, with the keepalive option set to true. If forceFetch is false and keepalive is not
+ * supported by the browser i.e. Firefox, the request is sent via the Beacon API.
  *
  * @param url URL that the request will be sent to
  * @param body the EventPayload object
+ * @param forceFetch if true, it will force the browser to call fetch instead of sendBeacon,
+ * regardless of whether keepAlive is supported
  */
-const post = (
```
AjayBenno, line 13: Thoughts on making a helper function which returns a boolean to use beacon or fetch? 

Then can unit test that based on the user agent etc

swong1213: Sounds good, I think that would be cleaner. I added a helper function in a new file (browser.ts).

src/post.ts
swong1213: Most resources I found suggested using the window.navigator.userAgent to determine the browser. I just added the user set browserAgent.browser as an additional check. Do you think this is sufficient? 

I did see online that using the window.navigator.userAgent is not always reliable because the userAgent can be easily spoofed (https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browsers). One suggestion for checking for Firefox is checking that the typeof InstallTrigger !== 'undefined', but I saw some comments that this isn't reliable as well. I also saw suggestions to use libraries e.g. https://github.com/lancedikson/bowser.

tests/post.test.ts
swong1213: Discussed with Merrick offline - mocking and then verifying the response is not really needed here as the main purpose of the tests are to check that it either calls fetch or sendBeacon.

src/post.ts
```diff
@@ -1,20 +1,25 @@
...
+    forceFetch: boolean,
 ) => {
-    return navigator.sendBeacon(url, JSON.stringify(body));
+    if (!forceFetch && useBeacon(body)) {
+        return navigator.sendBeacon(url, JSON.stringify(body));
+    }
+    return fetch(url, {method: 'POST', body: JSON.stringify(body), keepalive: true});
 }
 
 export default post;
```
ejaffee01, line 25: Just noticed this when working on implementing `report`. Is there a reason to export it the function at the bottom vs doing `export post = async (` ? I noticed that when going to import the method you do it like: 

`import post from "./post";`
instead of 
`import { merge } from './merge';`

I'm sure it does not make a big difference, but figured I'd check for consistency sake

swong1213: I think that when you export a default function (which means you're only exporting a single function from the file/module), you can import it like `import post from "./post"`. You can name the method however you like when you import it e.g. `import fetch from "./post"`, `import beacon from "./post"` would both import the same method. 

When you export named exports you can export multiple in one file/module, and you will have to import it like `import { merge, helper1, helper2 } from './merge` and the names need to match how they're defined in the file. However in your case I think you can just export a default function since you only have 1 function. I might change my PR to include both functions in 1 file.

src/post.ts
```diff
@@ -1,20 +1,32 @@
...
-export default post;
+/**
+ * Returns a boolean that determines whether the post method should use beacon. Firefox and Firefox
+ * for Android do not support fetch + keepAlive, so useBeacon will return true if the browser is
+ * Firefox or Firefox for Android.
+ *
+ * @param body the EventPayload object
+ */
+export function useBeacon(body: EventPayload) {
+    const agent = body.browserAgent;
```
mtian725, line 28: Should we window.navigator.userAgent? Or are we assuming this will be set from the report function? Someone needs to set the browserAgent somewhere.

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgent

swong1213: The useBeacon method checks for either the window.navigator.userAgent or the user-set browserAgent.browser. Would you suggest just using window.navigator.userAgent and ignoring the user-set value since we don't want it to override the window.navigator.userAgent?

AjayBenno: My original question for this was what cases would we want to be using the user set parameter to figure out if we should use beacon or fetch.

The main reason we need this check is because we want to make sure we are using the right function for the given browser right? Does the user set one tell us anything about the current browser?

swong1213: Hmm, don't think the user set parameter is reliable for checking the current browser and shouldn't take precedence over the window.navigator.userAgent. Removed the check for the user set parameter.

src/post.ts
```diff
@@ -1,20 +1,32 @@
...
 
-export default post;
+/**
+ * Returns a boolean that determines whether the post method should use beacon. Firefox and Firefox
+ * for Android do not support fetch + keepAlive, so useBeacon will return true if the browser is
+ * Firefox or Firefox for Android.
+ *
+ * @param body the EventPayload object
+ */
+export function useBeacon(body: EventPayload) {
```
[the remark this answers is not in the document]
swong1213: Updated to pass force fetch

src/post.ts
```diff
@@ -1,20 +1,31 @@
...
-export default post;
+/**
+ * Returns a boolean that determines whether the post method should use beacon. Firefox and Firefox
+ * for Android do not support fetch + keepAlive, so useBeacon will return true if the browser is
+ * Firefox or Firefox for Android.
+ *
+ * @param body the EventPayload object
+ */
+export function useBeacon(body: EventPayload, forceFetch: boolean) {
+    const agent = body.browserAgent;
```
mtian725, line 28: We can remove this right?

swong1213: oh I forgot to remove it, just pushed a new commit

mtian725 (approved):

PR #61, opened 2023-08-25
EventsApi: Implement report

ejaffee01:
Implenment report function for Analytics Events SDK.

J=FUS-5873
R=abenno, mtian, sawong
TEST=unit

Wrote unit tests that mock the response of post and getOrSetupSessionId and all pass. The tests check that merging behavior is correct, and the arguments passed into the post call are correct.


api-extractor.json
```diff
@@ -45,7 +45,7 @@
    *
    * SUPPORTED TOKENS: <projectFolder>, <packageName>, <unscopedPackageName>
    */
-  "mainEntryPointFilePath": "<projectFolder>/lib/esm/index.d.ts",
+   "mainEntryPointFilePath": "<projectFolder>/lib/esm/src/index.d.ts",
```
ejaffee01, line 48: For some reason something related to this change caused build errors where it could not find the path. /lib/esm/index.d.ts. After searching for this file, I found it at "/lib/esm/src/index.d.ts

src/Region.ts
```diff
@@ -8,8 +8,8 @@ import { EnumOrString } from './EnumOrString';
  * @public
  */
 export enum RegionEnum {
```
ejaffee01, line 10: switched to lowercase to make url setting easier.

src/setupSessionId.ts
```diff
@@ -0,0 +1,25 @@
+import { ulid } from 'ulidx';
+
+export const SESSION_ID_KEY = 'yext_analytics_session_id';
```
ejaffee01, line 3: This entire file is copied from how chat (or the previous iteration of the sdk) was getting sessionId.

tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "forceConsistentCasingInFileNames": true,
     "declaration": true,
     "declarationMap": true,
-    "sourceMap": true
+    "sourceMap": true,
+    "resolveJsonModule": true,
```
ejaffee01, line 11: he reason I added this property is so in AnalyticsEventReporter I can import values from package.json:

import { name, version } from '../package.json';

It complains without this property in the config.

Importing from package.json makes setting clientSdk super clean.

src/AnalyticsEventReporter.ts
```diff
@@ -1,38 +1,78 @@
...
+        ? (finalPayload.clientSdk as Record<string, string>)[name] = version
+        : finalPayload.clientSdk = {
+          [name]: version,
+        };
+
+      finalPayload.authorization = this.config.key
+        ? 'KEY ' + this.config.key
+        : 'Bearer ' + this.config.bearer;
+
+      const res = await post(
```
AjayBenno, line 59: Thoughts on instead calling useBeacon() up here? And then based on that we can condition our response handling.

```
var useBeacon = useBeacon(...)

const res = await post(useBeacon,...)

if(useBeacon){
...
}
else {
// used fetch + keep-alive
}
```

we could even make two post methods `postWithBeacon` and `postWithFetch` so that its even less confusing in the handling

src/post.ts
```diff
@@ -24,7 +24,7 @@ export function post(url: string, body: EventPayload, forceFetch: boolean) {
  *
  * @param body the EventPayload object
  */
-export function useBeacon(body: EventPayload, forceFetch: boolean) {
-    // keepAlive is not supported in Firefox or Firefox for Android
-    return !forceFetch && navigator.userAgent && navigator.userAgent.toLowerCase().includes('firefox');
-}
+export function useBeacon(body: EventPayload, forceFetch: boolean | undefined) {
```
mtian725, line 27: Isn't this `boolean | string`?

tests/setupRequestUrl.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  it('should set url correctly for env: prod, region: eu', () => {
+    const resultUrl = setupRequestUrl(EnvironmentEnum.Production, RegionEnum.EU);
+    const expectedUrl = 'https://eu.yextevents.com/accounts/me/events';
+
+    expect(resultUrl).toEqual(expectedUrl);
+  });
+
+  it('should set url correctly for env: prod, region: us', () => {
```
mtian725, line 21: nit: should set url correctly for env: **_sbx_**, region: us

src/AnalyticsEventReporter.ts
```diff
@@ -1,38 +1,78 @@
...
-
     /**
      * @param config - necessary analytics config: Must provide one and only
      * one of API Key or Bearer Token.
-     * 
+     *
      * @param payload - (optional) desired event values to report
      */
     constructor(config: AnalyticsConfig, payload?: EventPayload) {
-        const apiKeyIsSet = config.key !== undefined;
```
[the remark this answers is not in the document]
mtian725: Did you use prettier?

src/AnalyticsConfig.ts
```diff
@@ -26,4 +27,10 @@ export interface AnalyticsConfig {
      * This generates a ULID to tie together events in a single browsing session.
      */
     sessionTrackingEnabled?: boolean;
-}
+    /**
+     * Used to force sending the POST request via fetch + keepalive.
```
mtian725, line 31: This isn't exactly correct. If the browser supports fetch + keep alive, it'll just use fetch + keepAlive. This flag has no effect on it. 

I would also specify which browsers this would be applicable to.

src/AnalyticsEventReporter.ts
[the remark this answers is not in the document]
ejaffee01: Likely because of a linter. Undoing this indentation leads to vscode complaining about the indentation on every line. I believe we have prettier installed in the project.

src/merge.ts
```diff
@@ -10,7 +10,7 @@ import { EventPayload } from './EventPayload';
  *  Merging a non-existing key adds the key/value.
  *  When the value is an object, the above is applied iteratively.
  */
-export const merge = (original: EventPayload, newValues: PartialPayload): EventPayload => {
+const merge = (original: PartialPayload, newValues: PartialPayload): EventPayload => {
   const copy = JSON.parse(JSON.stringify(original));
```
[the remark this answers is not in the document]
ejaffee01: Added a check for that at the top of merge. 

However in the report function I believe we would still need to check if original is undefined, since right now the Reporter is setup so that the payload object (in the constructor) is optional.

src/AnalyticsEventReporter.ts
```diff
@@ -1,38 +1,78 @@
...
+      const currentPayload = this.payload === undefined ? payload : merge(this.payload, payload);
+      return new AnalyticsEventReporter(this.config, currentPayload);
     }
 
-    report(payload: object): Promise<EventAPIResponse> | boolean {
-        return false;
+    public async report(newPayload?: PartialPayload): Promise<boolean | EventAPIResponse> {
+      const finalPayload = this.payload && newPayload
+        ? merge(this.payload, newPayload)
+        : this.payload ?? newPayload ?? function(){throw Error('EventPayload is empty');}();
```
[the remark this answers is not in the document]
ejaffee01: I had to do this to include throwing an error as a part of the ternary operator. However as merge empty handling has been revised this is no longer necessary.

src/AnalyticsEventReporter.ts
```diff
@@ -1,38 +1,78 @@
...
+
+      const res = await post(
+        setupRequestUrl(
+          this.config.env,
+          this.config.region),
+        finalPayload as EventPayload,
+        this.config.forceFetch);
+
+
+      if (typeof res === 'boolean') {
```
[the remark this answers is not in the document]
ejaffee01: Not needed anymore

src/AnalyticsEventReporter.ts
```diff
@@ -1,38 +1,78 @@
...
+        return res;
+      } else if (!res.ok) {
```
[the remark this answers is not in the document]
ejaffee01: I believe we know its safe as we call await on the post call. Then if ok is undefined or does not have a property, it would be falsy, and the error case would execute.

src/AnalyticsEventReporter.ts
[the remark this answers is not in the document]
ejaffee01: > Did you use prettier?

I realized have eslint installed in vscode by default so I think there may be some additional formatting being done on top of prettier.

api-extractor.json
mtian725: Hmm, I see. I am not sure about what the right thing to do here is. I really don't think any config files should be changed. Let's loop back to this on the index PR since I have a feeling we can't really investigate why until all the pieces are together.

src/setupSessionId.ts
mtian725, line 3: Do we need to export the SESSION_ID_KEY? I'm not seeing it being used anywhere else.

src/setupRequestUrl.ts
```diff
@@ -0,0 +1,7 @@
+import { Environment, EnvironmentEnum } from './Environment';
+import { Region } from './Region';
+
+export function setupRequestUrl(env?: Environment, region?: Region): string {
+  const urlBase = 'yextevents.com/accounts/me/events';
```
mtian725, line 5: nit: let's move urlBase out of the function as a global variable.

src/AnalyticsEventReporter.ts
```diff
@@ -1,38 +1,78 @@
...
-        const currentPayload = this.payload === undefined ? payload : merge(this.payload, payload);
-        return new AnalyticsEventReporter(this.config, currentPayload);
+      const currentPayload = this.payload === undefined ? payload : merge(this.payload, payload);
+      return new AnalyticsEventReporter(this.config, currentPayload);
     }
 
-    report(payload: object): Promise<EventAPIResponse> | boolean {
-        return false;
+    public async report(newPayload: PartialPayload): Promise<boolean | EventAPIResponse> {
+      const finalPayload = this.payload === undefined ? newPayload : merge(this.payload, newPayload);
```
mtian725, line 38: Thoughts on using just `merge(this.payload, newPayload)`? Since `merge()` now just returns one of the arguments if either is empty, we can just edit the logic a bit further to also have undefined/null be equivalent to an empty object.

src/EventPayload.ts
```diff
@@ -8,6 +8,8 @@ import { Action } from './Action';
 export interface EventPayload {
   /** The user action which caused the event, e.g. ADD_TO_CART or THUMBS_UP  */
   action: Action;
+  /** The authorization token for the request. Formatted like KEY <api-key> or Bearer <bearer-token> */
```
mtian725, line 11: nit: add a note saying that this will be set up using the auth token passed into the config

src/setupSessionId.ts
ejaffee01: Removed `export`

src/AnalyticsEventReporter.ts
ejaffee01: Discussed offline. We will handle the case of newPayload or original being empty in the merge function itself. In report, we will pass an empty record if its null or undefined.

tests/AnalyticsEventReporter.test.ts
```diff
@@ -1,34 +1,217 @@
...
+        bearer: 'bearerToken',
+        sessionTrackingEnabled: true,
+      };
+      const reporter = new AnalyticsEventReporter(config).with({
+        action: 'ADD_TO_CART',
+        referrerUrl: 'https://yext.com',
+        count: 5,
+      });
+
+      const res = await reporter.report({
```
mtian725, line 186: Can we add the following tests:

- calling report with no argument
- Create then calling report with argument (no with() call)
- create then calling report without argument (no with() call)
- create and call with() more than once, and verifying that the two different Service objects are called with different values.

tests/AnalyticsEventReporter.test.ts
```diff
@@ -1,34 +1,395 @@
...
+      const reporter2 = reporter.with({
+        action: 'APPLY',
+        authorization: 'Bearer shouldNotUpdate',
+        destinationUrl: 'https://google.com',
+        clientSdk: {
+          chat: '1.0.1.0',
+        }
+      });
+
+      const res = await reporter2.report();
```
mtian725, line 373: Let's have another report() call for reporter1, and verify that the results it's been called with.

tests/AnalyticsEventReporter.test.ts
```diff
@@ -1,34 +1,395 @@
...
+
+      const config: AnalyticsConfig = {
+        key: 'validKey',
+        region: RegionEnum.EU,
+        forceFetch: false,
+      };
+
+      const service: AnalyticsEventService = new AnalyticsEventReporter(config);
+
+      const res = await service.report({action: 'ADD_TO_CART',
```
mtian725, line 270: nit: formatting?

src/AnalyticsEventReporter.ts
```diff
@@ -1,38 +1,78 @@
...
+      finalPayload.authorization = this.config.key
+        ? 'KEY ' + this.config.key
+        : 'Bearer ' + this.config.bearer;
+
+      const shouldUseBeacon = useBeacon(finalPayload as EventPayload, this.config.forceFetch);
+      const requestUrl = setupRequestUrl(this.config.env, this.config.region);
+
+      // If useBeacon returns true, return boolean response of postWithBeacon
+      if (shouldUseBeacon) {
+        return await postWithBeacon(requestUrl, finalPayload as EventPayload);
```
mtian725, line 62: nit: no need to cast `finalPayload as EventPayload`

src/AnalyticsEventReporter.ts
```diff
@@ -1,38 +1,78 @@
...
+      }
+
+      /** If useBeacon returns false, use postWithFetch.
+          If result is successful, return result json.
+          If request fails, return errors. */
+      const res = await postWithFetch(requestUrl, finalPayload as EventPayload);
```
mtian725, line 68: nit: no need to cast finalPayload as EventPayload

src/AnalyticsEventReporter.ts
```diff
@@ -1,38 +1,78 @@
...
+        ? (finalPayload.clientSdk as Record<string, string>)[name] = version
+        : finalPayload.clientSdk = {
+          [name]: version,
+        };
+
+      finalPayload.authorization = this.config.key
+        ? 'KEY ' + this.config.key
+        : 'Bearer ' + this.config.bearer;
+
+      const shouldUseBeacon = useBeacon(finalPayload as EventPayload, this.config.forceFetch);
```
mtian725, line 57: nit: here as well

mtian725 (approved):

PR #62, opened 2023-08-25
EventsApi: Implement index.ts

ejaffee01:
Implement index.ts by exporting an analytics object that takes in
a config, creates a reporter, and returns a service. This is the only
thing we are exporting in the sdk.

J=FUS-5974
R=mtian, abenno
TEST=manual

Tested by updating test-site to use new SDK.


test-site/src/index.ts
```diff
@@ -1,90 +1,8 @@
-import {CtaClick, provideAnalytics, providePagesAnalytics, provideChatAnalytics} from '@yext/analytics';
-import { CookieManager, provideConversionTrackingAnalytics } from '../../src';
+import analytics from '../../src/index';
```
AjayBenno, line 1: I think the import from `'@yext/analytics'` was intentional as this is meant to simulate how a user would interact with the SDK.

Is this import you used functionally the same?

AjayBenno (commented):
Your testing description didn't cover it but can you run the test site and confirm that is still works with these changes?

src/index.ts
```diff
@@ -1 +1,13 @@
...
+
+/**
+ * The Yext Analytics Events SDK
+ * @beta
+ */
+const analytics = (config: AnalyticsConfig): AnalyticsEventService => {
+  return new AnalyticsEventReporter(config);
+};
+
+export default analytics;
```
mtian725, line 13: I think we also want to export the AnalyticsConfig and AnalyticsEventService. Seeing warnings about them in the auto generated doc files `etc/analytics.api.md`. Plus, that's what was done before.

https://github.com/yext/analytics/blob/main/src/index.ts#L88

mtian725 (approved):

PR #63, opened 2023-08-30
Update test site

ejaffee01:
Update test site to use new version of SDK. In testing, this revealed some changes that needed to be made to the main package. These include updating the type of `entityId` in `EventPayload` to correctly match the Events Schema. This change also adds the Content-Type header to `postWithFetch` and updates associated tests. Lastly, this updates how we set clientSdk as using the package name is not supported since it includes special characters. 

To use the test-site:
- Run `npm i` in the main package
- Run `npm run build` in the main package
- Run `npm i` in test-site
- Run `npm run build` in test-site
- Run `npm run serve` in test-site

### DIFF
--- a/test-site/src/index.ts
+++ b/test-site/src/index.ts
@@ -1,19 +1,110 @@
-import {analytics} from '@yext/analytics';
+import { analytics } from '@yext/analytics';
 
 
-const chat = analytics({
+const analyticsProvider = analytics({
   key: process.env.YEXT_API_KEY,
+  sessionTrackingEnabled: false
+}).with({
+  action: "CHAT_LINK_CLICK",
+  pageUrl: "http://www.yext-test-pageurl.com",
+  destinationUrl: "http://www.yext-test-destinationurl.com",
+  referrerUrl: "http://www.yext-test-referrerurl.com",
+  label: "test-label",
+  timestamp: new Date(),
+  bot: false,
+  browserAgent: {
+    browser: "test-browser",
+    browserVersion: "test-browser-version",
+    os: "test-os",
+    osVersion: "test-os-version",
+    device: "test-device",
+    deviceClass: "test-device-class",
+    userAgent: "test-user-agent",
+  },
+  clientSdk: {
+    "testsdk": "version",
+  },
+  internalUser: false,
+  chat: {
+    botId: "analytics-test-bot"
+  },
+    count: 1,
+  customTags: {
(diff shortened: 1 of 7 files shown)


test-site/src/index.ts
```diff
@@ -1,19 +1,110 @@
...
+  },
+  visitor: {
+    "test-id-method": "visitor-test-id",
+  },
+});
+
+
+const analyticsProvideWithSessionTracking = analytics({
+  key: process.env.YEXT_API_KEY,
+  sessionTrackingEnabled: true
```
mtian725, line 51: no need to enable it, it should be true by defaul

test-site/src/index.ts
```diff
@@ -1,19 +1,110 @@
...
-    chat: {
-      botId: "analytics-test-bot"
-    }
-  })
-}
\ No newline at end of file
+  analyticsProvider.report();
+}
+
+export function fireCallToActionEvent() {
```
mtian725, line 97: nit: let's remove either fireCallToActionEvent() or fireSearchBarImpressionEvent()

test-site/src/index.ts
```diff
@@ -1,19 +1,105 @@
-import {analytics} from '@yext/analytics';
+import { analytics } from '@yext/analytics';
```
mtian725, line 1: One last thing, can we add a comment describing the weird behavior we saw in firefox
https://yext.slack.com/archives/G8UMH43HP/p1693419787243609?thread_ts=1693411231.167479&cid=G8UMH43HP

test-site/src/index.ts
```diff
@@ -1,19 +1,112 @@
...
+  analyticsProvider.report();
+}
+
+export function fireCallToActionEvent() {
+  analyticsProvider.report({action: "CALL_TO_ACTION"});
+}
+
+export function fireEventWithSessionTracking() {
+  analyticsProvideWithSessionTracking.report({
+    sessionId: "12345"
```
mtian725, line 109: I just realized that session ID should not be included in the payload if session tracking is off. It should be set to undefined even if sessionID is passed down. We should update the config comment as well.

mtian725, line 109: Because session tracking is enabled, we shouldnt need to pass anything down.

ejaffee01: Technically they can manually set sessionId if they want. But I'll remove it so this test auto generates it.

mtian725 (approved):

PR #65, opened 2023-08-31
Set up Documentation

mtian725:
Updated comments, tags, and exports to make sure everything is
surfaced for both the user and API generator
    
Also added README.md based off
https://yext.slack.com/archives/G01SSAYJVM4/p1693347938499419
    
J=FUS-5875
TEST=manual
    
ran npm run build and verified that the docs/ and etc/ was correct


README.md
```diff
@@ -27,149 +28,171 @@ First, install the library via [npm](https://www.npmjs.com/get-npm):
...
+  customTags: {
+      "promoCode": "SPRING15OFF"
+  },
+  customValues: {
+      "promoDiscount": 41.76
+  }
+});
 ```
+### Debugging
+We use `fetch()` + `keepalive` by default in [supported browsers](https://developer.mozilla.org/en-US/docs/Web/API/fetch) to make debugging easier. For browsers like Firefox that do not support `keepalive`, we use `fetch` only. Be warned, since `forceFetch` uses `fetch()` without `keepalive`, **requests in progress will be canceled if the page is unloaded**.
```
AjayBenno, line 124: This seems to indicate that we don't use beacon at all?

docs/analytics.eventapiresponse.md
```diff
@@ -0,0 +1,21 @@
+<!-- Do not edit this file. It is automatically generated by API Documenter. -->
+
+[Home](./index.md) &gt; [@yext/analytics](./analytics.md) &gt; [EventAPIResponse](./analytics.eventapiresponse.md)
+
+## EventAPIResponse interface
+
+Response from an analytic request.
```
AjayBenno, line 7: analytics

docs/analytics.eventpayload.entity.md
```diff
@@ -0,0 +1,17 @@
+<!-- Do not edit this file. It is automatically generated by API Documenter. -->
+
+[Home](./index.md) &gt; [@yext/analytics](./analytics.md) &gt; [EventPayload](./analytics.eventpayload.md) &gt; [entity](./analytics.eventpayload.entity.md)
+
+## EventPayload.entity property
+
+The Yext entity to which the event corresponds.
```
AjayBenno, line 7: Can we describe what the entityId vs entityUID is?

README.md
```diff
@@ -27,149 +28,171 @@ First, install the library via [npm](https://www.npmjs.com/get-npm):
...
-To use the library with Node, use the following import instead:
+You can add a `.with()` method to the root analytics service you initialized, which returns a new analytics object with the specified JSON merged on top of the existing defaults.
 ```ts
-const { provideSearchAnalytics } = require('@yext/analytics');
-``` 
+import { Analytics } from "@yext/analytics";
 
-Now that the search analytics reporter is initialized, let's fire off an event:
+// Root analytics service with no defaults.
+const rootAnalytics = new Analytics("MY_API_KEY");
```
ejaffee01, line 47: Can you do this? I thought you would need to pass it in like an (AnalyticsConfig) object. 

```
const rootAnalytics = new Analytics( { key: "MY_API_KEY" } );

File: README.md [hunk 5]
```diff
@@ -27,149 +28,171 @@ First, install the library via [npm](https://www.npmjs.com/get-npm):
...
-  verticalKey: 'people',
-  searcher: 'VERTICAL',
-  queryId: '12345678-1234-1234-1234-123456789012'
+chatAnalytics.report({
+  action: "CHAT_IMPRESSION"
 });
 ```
+### Additional Configuration 
+#### Session Tracking
+Session tracking is now available for Chat, Pages, and Search. Yext uses a browser-based method (sessionStorage) to track this. By default, session tracking is **enabled** in **both** the US and EU environments. For users whose Yext accounts are located in the EU environment, this can be disabled by setting `sessionTrackingEnabled` to `false`.
```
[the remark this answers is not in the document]
cxi2319: Sgtm, let's remove that and make it `This can be disabled...`

README.md
mtian725: Ah, you're right.

README.md
mtian725: Ah, I forgot chris edited it since we were still figuring out whether we will still use beacon

docs/analytics.eventpayload.entity.md
mtian725: I'll just rebase off your change once it's up

ejaffee01: Use this when rebasing to avoid having others changes show up in your diff. 

https://stackoverflow.com/questions/18323665/pull-request-on-github-showing-commits-rebased-from-master

README.md
```diff
@@ -14,9 +14,10 @@ A Typescript library for sending Yext Analytics events.
 
 [Full Documentation](./docs/analytics.md)
```
ejaffee01, line 15: Sort of unrelated to the generated docs, but we should probably update this too: 
Screenshot 2023-08-31 at 12 38 53 PM

src/Environment.ts
```diff
@@ -9,6 +9,7 @@ import { EnumOrString } from './EnumOrString';
  */
 export enum EnvironmentEnum {
   Production = 'PRODUCTION',
+  /** unsupported in the EU region */
```
ejaffee01, line 12: I forgot sbx was not supported in the EU. Should we update setupRequestUrl to throw an error if env = sbx and region = eu is passed in?

README.md
mtian725: I don't think either of us have permission to edit it. There would be a gear icon like so if we did.

Screenshot 2023-08-31 at 2 01 15 PM

src/Environment.ts
mtian725: Yeah, can edit it to throw an error

mtian725: Actually, I'll put it up as a different PR just so the it's reviewed separately since we'll probably need to edit the test files.

ejaffee01 (approved):
LGTM

AjayBenno (approved):

PR #74, opened 2023-09-12
EventsSdk: Update report return type

ejaffee01:
Remove errors from EventApiResponse. Instead we will report errors by calling Promise.reject() which will print out the errors from the events api. If it is successful, we return (but do not print) the result json with the event id if called from fetch, and a boolean as a string if called from post.

J=FUS-5895
R=mtian, abenno
TEST=manual

Made valid and invalid requests to the test site and ensured expected behavior.


ejaffee01:
Code coverage test started failing again - but not sure why since I did not really remove any function calls. I think we are just hovering right around the 80% threshold

src/AnalyticsEventReporter.ts
```diff
@@ -60,22 +60,19 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
       const shouldUseBeacon = useBeacon(finalPayload, this.config.forceFetch);
       const requestUrl = setupRequestUrl(this.config.env, this.config.region);
 
-      // If useBeacon returns true, return boolean response of postWithBeacon
+      // If useBeacon returns true, return boolean response of postWithBeacon as string.
       if (shouldUseBeacon) {
-        return await postWithBeacon(requestUrl, finalPayload);
+        return String(postWithBeacon(requestUrl, finalPayload));
```
AjayBenno, line 65: In this case if postFails, shouldn't we do Promise.reject instead? Can do Promise.reject("Beacon Call Failed")

If it's successful, I think can just do Promise.resolve("");

We should also wrap the postWithBeacon call within a promise, otherwise it won't be async as well. Something like 
```
return new Promise((resolve), (reject) => {
 if(postWithBeacon){
  resolve("");
}else{
  reject("Failed Beacon Call")
}
```

src/AnalyticsEventReporter.ts
```diff
@@ -60,22 +60,19 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
...
       }
 
       /** If useBeacon returns false, use postWithFetch.
           If result is successful, return result json.
           If request fails, return errors. */
       const res = await postWithFetch(requestUrl, finalPayload);
```
AjayBenno, line 71: I realized that this function doesn't really send the data asynchronously as we are doing `await` on the post call.

Instead can we try to chain the promises so that we just return the chained promise. Something like 

```
const postPromise = postWithFetch(..)
postPromise.then((response) => {
  response.json()['id']
})
.catch(...)
```

That way the post report call is truly async

src/EventAPIResponse.ts
```diff
@@ -8,8 +8,4 @@ export interface EventAPIResponse {
    * Generated ID for the event.
```
AjayBenno, line 8: Since we're just returning a string do we need this file anymore?

src/AnalyticsEventReporter.ts
```diff
@@ -60,22 +59,22 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
...
-        let errorMessage = `Events API responded with ${res?.status}: ${res?.statusText}`;
-        body?.errors?.forEach(e => errorMessage += `\nError: ${e}.`);
-        throw new Error(errorMessage);
-      }
-      const resJson = await res?.json();
-      return resJson;
+      If result is successful, return result json.
+      If request fails, return errors. */
+      return postWithFetch(requestUrl, finalPayload)
+        .then((response) => response)
```
AjayBenno, line 77: Do we not need to do response.text() or response.json?

tests/AnalyticsEventReporter.test.ts
```diff
@@ -359,27 +310,11 @@ describe('Test report function', () => {
...
-        headers: new Headers(),
-        body: null,
-        bodyUsed: false,
-        arrayBuffer: jest.fn(),
-        blob: jest.fn(),
-        formData: jest.fn(),
-        text: jest.fn()
-      });
+      mockPostWithBeacon.mockReturnValue(false);
+      mockPostWithFetch.mockResolvedValue({errors: ['Bad Request']});
```
AjayBenno, line 314: Do we ever mock the failed promise? i only see us mockPostWithFetch with mockResolvedValue here in this file.

src/AnalyticsEventReporter.ts
ejaffee01: I do not believe so. I just tried using .text() and .json and both result in undefined being returned.

tests/AnalyticsEventReporter.test.ts
ejaffee01: Ah these should be using mockRejectedValue

AjayBenno (approved):
LGTM

PR #75, opened 2023-09-12
EventsSdk: Add CDN actions

ejaffee01:
Add missing actions for deploying sdk to CDN.
- build_and_deploy_hold
- extract_versions
- should_deploy_major_version

J=FUS-5906
R=mtian, abenno
TEST=manual

Validated yaml

### DIFF
--- a/.github/workflows/build_and_deploy_hold.yml
+++ b/.github/workflows/build_and_deploy_hold.yml
@@ -1,0 +1,71 @@
+name: Build and deploy with a hold state
+
+on:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  call_build_US:
+    uses: ./.github/workflows/build.yml
+    with:
+      region: US
+
+  call_build_EU:
+    uses: ./.github/workflows/build.yml
+    with:
+      region: EU
+
+  call_extract_versions:
+    uses: ./.github/workflows/extract_versions.yml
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_deploy_hold.yml
```diff
@@ -0,0 +1,71 @@
+name: Build and deploy with a hold state
+
+on:
+  push:
+    tags:
```
AjayBenno, line 5: What does this mean? What do we need to do to get this workflow to kick off?

ejaffee01: How its currently setup I think this will run automatically on any push. I think we'd want to change it so its to every push to main? 
```
on:
  push:
    branches:
      - main
```
Could also require a manual run: 
```
on:
  workflow_call:
```

ejaffee01:
Here's how this works:

1. build_and_deploy_hold.yml is called when a new version tag is added. (can also be manually triggered)
2. This calls build.yml which uses vite to generate a build-output-<region> file
3. calls extract_versions and should_deploy_major_version to determine the version and what needs to be deployed. 
4. calls deploy_hold which deploys to aws/gcp using secret keys.

mtian725 (approved):
Let's see if this works.

PR #77, opened 2023-09-13
EventsSDK: Remove unused package dependencies

mtian725:
Removed all unused dependencies in package.json that were no longer being used after the changes made in the latest beta build. 

J=FUS-5896
TEST=manual

Ran `npx depcheck` and uninstalled whatever was listed. Only exception being `jest-fetch-mock` which is being used by post.test.ts but was marked as unused.

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -1,35 +1,5 @@
 The following NPM package may be included in this product:
 
- - cross-fetch@3.1.5
-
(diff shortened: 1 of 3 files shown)


ejaffee01:
Did we want to try to add the result of `npm depcheck` to the output of the build command?

ejaffee01 (commented):
Did we want to try to add the result of npm depcheck to the output of the build command?

AjayBenno (approved):

PR #78, opened 2023-09-14
EventsSDK: Create New Playwright GH Action

mtian725:
Ran `npm init playwright@latest` and changed branch pattern matching to `**` so action is ran for all branches.

J=FUS-5879
TEST=auto

Pushed change and verified that new action is listed

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ '**' ]
+  pull_request:
+    branches: [ '**' ]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+    - name: Install dependencies
+      run: npm ci
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
+    - name: Run Playwright tests
+      run: npx playwright test
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ temp/
 .env
 .vscode/
 .idea/
+/test-results/
+/playwright-report/
+/playwright/.cache/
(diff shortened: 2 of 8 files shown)


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
+    - name: Run Playwright tests
+      run: npx playwright test
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
```
AjayBenno, line 27: what is retention-days 30 mean?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ '**' ]
+  pull_request:
+    branches: [ '**' ]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
```
AjayBenno, line 10: Do you know if this machine costs any different?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ '**' ]
+  pull_request:
+    branches: [ '**' ]
+jobs:
+  test:
+    timeout-minutes: 60
```
AjayBenno, line 9: 60 minutes seems pretty long, any reason this can't be like 5?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ '**' ]
```
AjayBenno, line 4: nit: think we can update to https://github.com/yext/analytics/blob/main/.github/workflows/run_tests.yml so that it runs on everything?

ejaffee01 (commented):
Maybe to make sure this works we could add a small test that opens up a browser? Though I'm not sure if this is possible as iirc actions have to be in main to be ran. Did see this for running github actions locally but not sure if its worth it in this case https://github.com/nektos/act.

.github/workflows/playwright.yml
mtian725: We can change this. Most of this is the default provided by running the command.

PR #81, opened 2023-09-14
EventsSDK: Update v0 to use keepalive

ejaffee01:
Users with v0.6.4 were experiencing requests erroring when users navigate away from the page. This adds the use of the keepalive flag which should prevent this error in browsers that support it.

J=FUS-5933
R=mtian, abenno
TEST=manual

Ran test site

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/analytics",
-  "version": "0.6.4",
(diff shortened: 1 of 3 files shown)


AjayBenno (commented):
Can we also bump the version to v0.6.5 as well?

AjayBenno (approved):

PR #83, opened 2023-09-14
EventsSDK: Create New Playwright GH Action

mtian725:
Ran npm init playwright@latest and edited the github action file to suit our needs.

J=FUS-5879
TEST=auto

Pushed change and verified that new action is listed. Also, verified that the `run_tests` action only runs `npx jest`

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,23 @@
+name: Playwright Tests
+on: [push, pull_request]
+jobs:
+  test:
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+    - name: Install dependencies
+      run: npm ci
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
+    - name: Run Playwright tests
+      run: npx playwright test
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
--- a/package.json
+++ b/package.json
@@ -33,6 +33,8 @@
     "@babel/preset-typescript": "^7.14.5",
     "@microsoft/api-documenter": "^7.15.3",
     "@microsoft/api-extractor": "^7.19.4",
+    "@playwright/test": "^1.38.0",
+    "@types/node": "^20.6.0",
     "@typescript-eslint/eslint-plugin": "^4.28.2",
     "@typescript-eslint/parser": "^4.28.2",
     "eslint": "^7.30.0",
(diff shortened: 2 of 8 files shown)


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,23 @@
+name: Playwright Tests
+on: [push, pull_request]
+jobs:
+  test:
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
```
mtian725, line 6: Do you know if this machine costs any different?

https://github.com/yext/analytics/pull/78/files/43fe152ffcf9a9768f99d1c39b764777f6805c56#r1326142457

mtian725: That's what it was defaulted as, and that's what deploy_hold.yml uses. For clarity, what do you mean by costs?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,23 @@
...
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
+    - name: Run Playwright tests
+      run: npx playwright test
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
```
mtian725, line 23: what is retention-days 30 mean? (copied from old PR) @AjayBenno 

https://github.com/yext/analytics/pull/78/files/43fe152ffcf9a9768f99d1c39b764777f6805c56#r1326142207

mtian725: I think it means how long the artifact produced by the run is retained. I'm not seeing anything online that would give me an exact definition. If this is the case, 30 days is definitely overkill. Thoughts on 2 or 3 days?

.github/workflows/playwright.yml
AjayBenno: Cool, was just curious.

.github/workflows/playwright.yml
AjayBenno: Apparently it doesn't really cost us anymore so 30 seems fine for now.

package.json
```diff
@@ -33,6 +33,8 @@
     "@babel/preset-typescript": "^7.14.5",
```
AjayBenno, line 33: Curious does the test command run the playwright tests? or can we add a playwright one? Just so npm run tests ends up actually running both playwright and npm tests.

mtian725: Nah, it just runs the jest tests. To run the playwright ones we would need to run `npx playwright test`. The suite of tests provided by the init script are the following

```

  npx playwright test
    Runs the end-to-end tests.

  npx playwright test --ui
    Starts the interactive UI mode.

  npx playwright test --project=chromium
    Runs the tests only on Desktop Chrome.

  npx playwright test example
    Runs the tests in a specific file.

  npx playwright test --debug
    Runs the tests in debug mode.

  npx playwright codegen

mtian725: As discussed offline, the test command runs both jest and playwright. Actions that currently call npm run test should be changed to just be running the unit tests

mtian725: As discussed offline, realized that we can pass down inputs to the reusable workflows. 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idwith

AjayBenno (approved):
LGTM

PR #86, opened 2023-09-15
Use Commit Hook to Auto Format Code

alinayh:
Use Husky to create pre-commit hook
Format the entire repo in this pull request
Note: People need to run "npm run prepare" to use it.
The code change is on .husky/pre-commit, .eslintignore, .prettierignore, and package.json. Other changes are made by prettier and eslint.

J=FUS-5893
TEST=auto

Successfully run prettier and eslint when I run "git commit"

PR #88, opened 2023-09-15
Use Commit Hook to Auto Format Code

alinayh:
Use Husky to create pre-commit hook

Note: People need to run "npm run prepare" to use it.
I'll create another PR to format the entire codebase after everyone merges their PR to unified-sdk.

J=FUS-5893
TEST=auto


.husky/pre-commit
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
```
AjayBenno, line 2: what does this line do?

package.json
```diff
@@ -20,7 +20,9 @@
   ],
   "scripts": {
     "build": "rm -rf lib/** && tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json && npm run api-extractor && npm run generate-docs && npm run generate-notices && vite build",
-    "lint": "eslint .",
+    "prepare": "husky install",
```
AjayBenno, line 23: Does this mean everyone installing the library needs to run npm prepare in order for the commit hooks to work?

Didn't quite follow what you meant by you need to run prepare.

alinayh: In order to use commit hook, it’s required to run ‘husky install’. Just install the library is not enough.

.husky/pre-commit
alinayh: After running 'husky install', it'll add husky.sh under .husky/_/. This is to find the husky configuration file. (Seeing chat does this way.

mtian725: @alinayh Can you provide a link to the chat example?

alinayh: actually seeing this in the pages repo: https://github.com/yext/pages/blob/main/.husky/pre-commit

alinayh:
The failed checks are expected, as I added a format-check into build. Before formatting the world is done, they'll fail.

package.json
```diff
@@ -19,8 +19,10 @@
     "THIRD-PARTY-NOTICES"
   ],
   "scripts": {
-    "build": "rm -rf lib/** && tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json && npm run api-extractor && npm run generate-docs && npm run generate-notices && vite build",
-    "lint": "eslint .",
+    "build": "rm -rf lib/** && tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json && npm run format-check && npm run api-extractor && npm run generate-docs && npm run generate-notices && vite build",
+    "prepare": "husky install",
+    "format-check": "prettier --check . && eslint .",
+    "format-fix": "prettier --write . && eslint . --fix",
```
mtian725, line 25: Thoughts on changing `format-fix` to just `clean`?

package.json
mtian725: I don't understand. If the commit hook is working, then the entire repo should be formatted, and we should be seeing much more being changed in this CR. 

Additionally, the build is failing due to the formatting being off, which this commit cook is supposed to fix before it's merged.

package.json
```diff
@@ -35,11 +37,13 @@
     "@microsoft/api-extractor": "^7.19.4",
     "@typescript-eslint/eslint-plugin": "^4.28.2",
     "@typescript-eslint/parser": "^4.28.2",
-    "eslint": "^7.30.0",
+    "eslint": "^7.32.0",
```
mtian725, line 40: Out of curiosity, is there a reason why the version went up?

package.json
alinayh: See here: https://github.com/yext/analytics/pull/86

package.json
alinayh: I ran `npm install prettier eslint husky --save-dev` to install packages and did not notice eslint has already installed. But it should not matter in terms of behavior.

package.json
alinayh: I found format-fix is a better name than clean, since we also have format-check. Do you have a reason for changing? `clean` does not look obvious to me on what it does at the first glance.

mtian725: Nah, just thinking about it. What is there is fine.

package.json
mtian725: So is this just to add the hook? And another is to just reformat the world?

alinayh: yes

PR #97, opened 2023-09-28
EventsSDK: Add Playwright Tests

ejaffee01:
Add playwright tests to serve as acceptance tests for this SDK. They will run the test site and click the buttons on three browsers, Chromium, Firefox, and Webkit and check for a 202 response. They can be run with the other unit tests using npm run test, or on their own with npx playwright test. The browsers can also be viewed live as they execute by using npx playwright test --headed.

J=FUS-5909
R=mtian, sawong, abenno
TEST=auto


.gitignore
```diff
@@ -9,3 +9,4 @@ temp/
 /test-results/
 /playwright-report/
 /playwright/.cache/
+**/.DS_Store
```
mtian725, line 12: why was this removed?

.github/workflows/playwright.yml
```diff
@@ -1,23 +1,27 @@
...
-      run: npx playwright install --with-deps
+      run: npm ci && npm i --prefix ./test-site
+    - name: build package
+      run: npm run build
     - name: Run Playwright tests
+      env:
+        # Refer to: https://github.com/microsoft/playwright/issues/6500#issuecomment-1577221956
+        HOME: /root
       run: npx playwright test
-    - uses: actions/upload-artifact@v3
```
mtian725, line 18: Shouldn't we keep this? I thought the point of this was the save the test demo of each run.

.gitignore
ejaffee01: I accidentally committed it once, I do not think it needs to be in the repo (its not currently on main or dev/unified-sdk. To avoid this happening again I added to .gitignore.

.github/workflows/playwright.yml
ejaffee01: It was not doing anything as we do not generate a playwright-report directory on run of the playwright tests, so I removed. We have it set to print out the report in the logs, or generate an html page if running locally. I do not think having an artifact will be that useful, and will just slow the tests down.

AjayBenno (approved):
LGTM

PR #100, opened 2023-09-29
Add Git Blame Ignore File

alinayh:
Exclude changes for formatting the world.

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view


.git-blame-ignore-revs
```diff
@@ -0,0 +1,8 @@
+# Format the world using Prettier and ESLint
+f042747aa077cda1a01a2b72ae8178c3ac04416a
```
mtian725, line 2: Are these commit change ids? Why are there 6? I thought we just had one format the world change? Did we not squash it?

alinayh: oh yea I don't know we need to squash it before.

mtian725 (approved):

PR #108, opened 2023-10-05
EventsSDK: Merge main into dev/unified-sdk

ejaffee01:
dev/unified-sdk says that it is 6 commits behind main. To fix this, we need to merge main into dev/unified-sdk. However, there are merge conflicts which makes this easier to resolve locally. dev/merge-branches is my local state of main being merged into dev/unified-sdk. You can confirm this by trying to open a pull request comparing dev/merge-branches and main and GitHub will tell you that there is nothing to merge. 

Following this merge, dev/unified-sdk and main should be in the same state.


.github/workflows/playwright.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       with:
         node-version: 18
     - name: Install dependencies
-      run: npm ci
```
mtian725, line 19: Don't we want to keep this? We made these changes so local development is less confusion.

mtian725 (commented):
I'm a bit confused. The changes in this look like we are undoing changes we did

.github/workflows/playwright.yml
ejaffee01: Yeah you're right - this is not correct. I just fixed them and pushed, but now it says there are no differences between the branches. I think I will close this for now, and we can sort out the differences between main and dev/unified-sdk when we release the next version.

ejaffee01:
> I'm a bit confused. The changes in this look like we are undoing changes we did

Yea

PR #117, opened 2023-10-25
EventsSDK: Add missing properties to EventPayload & make alphabetical

ejaffee01:
Added missing properties to event payload: 

- value
- location
     - Required adding Coordinates type to represent that location can be an object containing latitude or longitude, or be a 
     country string. 
- nonce
- searchTerm

In addition rearranged the file to make props in alphabetical order. 

Testing: 
- Ran unit and playwright tests and all pass
- Looked at official events api schema side by side with this EventPayload and made sure no properties were missing.


src/EventPayload.ts
```diff
@@ -100,17 +60,73 @@ export interface EventPayload {
...
+  sites?: {
+    /* The UID of the site an event was tied to. */
+    siteUid?: number;
+    /* The ID of the template from which a site was generated. */
+    template?: string;
+  };
+  /** The timestamp at which the event occurred, in ISO format. */
+  timestamp?: Date | string;
+  /** The monetary value of the event. */
+  value?: {
```
mtian725, line 125: Double checking is this `value` field the only addition made?

src/Coordinates.ts
```diff
@@ -0,0 +1,4 @@
+export type Coordinates = {
```
mtian725, line 1: Why did we add this? You didn't explain why in your commit msg.

mtian725 (commented):
What did you do for testing?

src/EventPayload.ts
ejaffee01: Added additional additions to commit message.

src/Coordinates.ts
ejaffee01: Added to commit mesage.

mtian725 (commented):
Please update the test files as well. You said you unit and playright tests passed as well, but none of your changes were tested.

src/EventPayload.ts
```diff
@@ -100,17 +60,73 @@ export interface EventPayload {
...
+  label?: string;
+  /** The locale of the user who generated the event. */
+  locale?: string;
+  /** The location information of the visitor for the event.
+   * Either a Coordinates object with both latitude and longitude or a string
+   * with the country of the visitor for the event, as a ISO 3166-1 alpha-2 country code. */
+  location?: Coordinates | string;
+  /* Deduplication ID for the event.
+  Events from the same app for the same business with the same nonce will get de-duplicated. */
+  nonce?: string;
```
AjayBenno, line 89: We can drop nonce, we meant to clean this up from the backend but have not yet.

AjayBenno (commented):
LG comment about nonce

PR #118, opened 2023-11-06
EventsSDK: Support Google Tag Manager

ejaffee01:
In order to support Google Tag Manager, we will export a new function just specifically for sending Events API requests from it. This also adds a convertTypesGTM function which is used to convert booleans and numbers represented as strings to their correct types. This is unit tested in convertTypes.test.ts.

J=https://yexttest.atlassian.net/browse/FUS-6055
TEST=auto, unit


src/index.ts
```diff
@@ -19,3 +43,10 @@ export * from './EventPayload';
 export * from './EnumOrString';
 export * from './Action';
 export * from './VersionLabel';
+
+declare global {
```
ejaffee01, line 47: FYI: Without this global addition, I can not do `window['analyticsEventPayload'];` on line 20. It causes the error: 
`Element implicitly has an 'any' type because index expression is not of type 'number'.`

etc/analytics.api.md
```diff
@@ -26,6 +26,9 @@ export interface AnalyticsEventService {
     with(payload: EventPayload): AnalyticsEventService;
 }
 
+// @public
+export function analyticsGTM(): Promise<string>;
```
mtian725, line 30: Why do we need to create another function just for GTM?

src/convertTypesGTM.ts
```diff
@@ -0,0 +1,70 @@
...
+  'amount',
+  'latitude',
+  'longitude',
+  'timestamp',
+  'bot',
+  'internalUser',
+  'isDirectAnswer'
+];
+
+export function convertTypesGTM(data: Record<string, unknown>): EventPayload {
```
mtian725, line 18: What exactly are we converting?

src/index.ts
mtian725: Why do we need to save a payload to the `window` object in the first place?

etc/analytics.api.md
ejaffee01: Originally we tried to just use the original function, however google tag manager uses Sandboxed Javascript which has limited functionality. We can inject the script, but then we cannot call with because it returns a type not accepted by GTM, which results in undefined. After this we considered sending a post request directly to the events api instead, but that is not supported either (only get requests are). 

The workaround is for google tag manager to store the event payload in the window object as analyticsEventPayload. It will have two children, config and payload. Then GTM will call this function which reads from the window object and calls report.

src/convertTypesGTM.ts
ejaffee01: User input from GTM always comes in as strings. We need to convert string represented numbers and booleans to the correct type before sending the request. This is what that takes care of.

src/index.ts
ejaffee01: See other comment below.

src/index.ts
```diff
@@ -11,6 +12,29 @@ export function analytics(config: AnalyticsConfig): AnalyticsEventService {
   return new AnalyticsEventReporter(config);
 }
 
+/**
+ * The exported function of Yext Analytics Events SDK used for Google Tag Manager.
+ * @public
+ */
+export function analyticsGTM(): Promise<string> {
+  const payload = window['analyticsEventPayload'];
```
mtian725, line 20: Using `payload` as a name here is a bit confusing since we have a payload object, but this is a tuple of the EventConfig and EventPayload

src/convertTypesGTM.ts
mtian725: Then let's call this something like `convertStringtoValue` since GTM is not giving us some custom value that we are destructuring.

src/convertTypesGTM.ts
```diff
@@ -0,0 +1,70 @@
...
+    ...result,
+    action: result.action as EventPayload['action']
+  };
+}
+
+// Function to convert custom values within customValues object
+function convertCustomValues(obj: Record<string, unknown>) {
+  for (const key in obj) {
+    if (
+      obj.hasOwnProperty(key) &&
```
mtian725, line 63: Isn't `obj.hasOwnProperty(key)` always true?

src/index.ts
ejaffee01: renaming to gtmPayload to be more specific.

src/index.ts
```diff
@@ -11,6 +12,32 @@ export function analytics(config: AnalyticsConfig): AnalyticsEventService {
...
 
+/**
+ * The exported function of Yext Analytics Events SDK used for Google Tag Manager.
+ * @public
+ */
+export function analyticsGTM(): Promise<string> {
+  const gtmPayload = window['analyticsEventPayload'];
+  let response: Promise<string>;
+  if (gtmPayload) {
+    const config = gtmPayload[0][1] as Record<string, unknown>;
```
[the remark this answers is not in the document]
ejaffee01: We are guaranteed that gtmPayload will be length 2. Config will always be inserted, as GTM will not allow for an empty API Key, session tracking, etc. field. For payload, action is a required field so that will not be empty either.

src/convertStringToValue.ts
```diff
@@ -0,0 +1,70 @@
+import { EventPayload } from './EventPayload';
+
+// Define the list of possibly numerical properties to check and convert
+const propertiesToCheck = [
```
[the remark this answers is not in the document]
ejaffee01: I looked into this a little bit and it appears that since interfaces do not exist at runtime, extracting any information from it at run time is not possible. It seems in order to use `typeof` on fields of EventPayload the object would need to be initialized. 
I also considered just looping through the data object and just determining to convert based on if it is a string representing a boolean or a number, but I am concerned that this will cause more complexity for fields that are strings that look like numbers, but are meant to stay strings in the payload. For example, browserVersion and alike fields. 

I know its not ideal to maintain a hard coded list, but I can add this to the wiki for something to remember when adding a new field to the events schema.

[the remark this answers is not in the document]
AjayBenno: gotcha thanks for the explanation.

src/index.ts
```diff
@@ -11,6 +12,39 @@ export function analytics(config: AnalyticsConfig): AnalyticsEventService {
...
 }
 
+/**
+ * An alternative entry point for the Yext Analytics Events SDK, currently used by Google Tag Manager.
+ * This method reads the config and payload from the variable analyticsEventPayload stored
+ * in the global window object. The config and payload are then passed to the report function to be sent
+ * to the Yext Analytics Events API.
+ * @public
+ */
+export function reportBrowserAnalytics(): Promise<string> {
```
AjayBenno, line 22: Awesome, love the public doc here

tests/reportBrowserAnalytics.test.ts
```diff
@@ -0,0 +1,38 @@
+import { reportBrowserAnalytics } from '../src/index';
+import { AnalyticsEventReporter } from '../src/AnalyticsEventReporter'; // replace with actual import path
+
+describe('reportBrowserAnalytics', () => {
+  let reportSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    // Mock window['analyticsEventPayload']
+    (global as any).window = {};
```
AjayBenno, line 9: Just to confirm, nothing we can do about these warnings?

AjayBenno (approved):
LGTM, had a small question about the actions warnings but feel free to ship if its expected.

tests/reportBrowserAnalytics.test.ts
ejaffee01: Unfortunately not. The alternatives cause actual errors. 

- Changing to unknown leads to `Object is of type 'unknown'.`

- Changing to never leads to `Property 'window' does not exist on type 'never'.ts(2339)`

- Removing the cast complety leads to `Type '{}' is not assignable to type 'Window & typeof globalThis'.

If we decide this is noisy we can also add an ignore comment to the top of the file in a future PR. 
`/* eslint-disable @typescript-eslint/no-explicit-any */`

PR #123, opened 2024-01-24
EventsSDK: Remove Partial Payload

ejaffee01:
Remove PartialPayload in favor of making all EventPayload properties optional. 
This required refactoring merge.ts.

J=FUS-6203
R=abenno, mtian
TEST=auto

Run automated workflows (which required updating the path for the run_tests yml) and added a test to merge.test.ts that merges two payloads without an action property.


src/merge.ts
```diff
@@ -11,41 +11,35 @@ import { EventPayload, PartialPayload } from './EventPayload';
  *  When the value is an object, the above is applied iteratively.
```
ejaffee01, line 11: Note: Change to `recursively` on next commit.

mtian725 (approved):
LGTM

PR #124, opened 2024-01-25
EventsSDK: Revise Authorization in AnalyticsConfig

ejaffee01:
Previously we had to fields in AnalyticsConfig that we're optional, `key` and `bearer`, where we throw an error if none or both are set. This change updates the interface so it is more clear that only one is required. There is now `authorizationType` which is either a string `apiKey` or `Bearer` and `authorization` which is the token itself. We need `authorizationType` so we can append the proper prefix of `KEY ` or `Bearer ` to the final authorization property to avoid passing this work onto the user. This also updates necessary test files to reflect this change, and removes logic for ensuring that only one of key or bearer is set. 

[J=FUS-6201](https://yexttest.atlassian.net/browse/FUS-6201)
R=abenno, mtian, mkilpatrick
TEST=auto


src/AnalyticsConfig.ts
```diff
@@ -7,14 +7,10 @@ import { Region } from './Region';
...
-  /** The API Key or OAuth for accessing the Analytics Events API
-   * Only one of key or bearer should be set.
-   * */
-  key?: string;
-  /** The bearer token for accessing the Analytics Events API.
-   * Only one of key or bearer should be set.
-   * */
-  bearer?: string;
+  /** Used for specifying if an API Key or Bearer Token is used for the authorization property. */
+  authorizationType: 'apiKey' | 'Bearer';
```
mkilpatrick, line 11: I know I put this as the example, but do you want the two types here to have a consistent capitalization?

src/AnalyticsEventReporter.ts
```diff
@@ -18,13 +18,6 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
    * @param payload - (optional) desired event values to report
    */
   constructor(config: AnalyticsConfig, payload?: EventPayload) {
-    const apiKeyIsSet = config.key !== undefined;
-    const bearerTokenIsSet = config.bearer !== undefined;
-    const configIsValid =
-      (apiKeyIsSet || bearerTokenIsSet) && !(apiKeyIsSet && bearerTokenIsSet);
-    if (!configIsValid) {
```
mkilpatrick, line 25: I think you need to still validate that the authorization and authorizationKey are defined. Someone not using typescript won't get the type checking so could still not set the values at all.

mkilpatrick (approved):

PR #125, opened 2024-01-25
EventsSDK: Remove Enums

ejaffee01:
We learned that using enums in TypeScript is bad practice in favor of types. This updates `Environment`, `Region`, and `VersionLabel` files to use types instead of enums. In addition, we delete the EnumOrString type and update `setupRequestUrl` and relevant tests. 

[J=FUS-6202 ](https://yexttest.atlassian.net/browse/FUS-6202)
R=abenno, mtian
TEST=auto


mtian725 (commented):
Is the removal of the EnumOrString type an issue for existing users? Are there even existing users to worry about?

ejaffee01:
> Is the removal of the EnumOrString type an issue for existing users? Are there even existing users to worry about?

I don't think so. I think all existing users would be using the sdk from the script tag, which would point them to the older version still using EnumOrString.

src/Region.ts
```diff
@@ -20,4 +6,5 @@ export enum RegionEnum {
  *
  * @public
  */
-export type Region = EnumOrString<RegionEnum>;
+
+export type Region = 'us' | 'eu';
```
mtian725, line 10: Not sure how important this is, but we should probably be consistent regarding whether we want this to be upper or lowercase. Right now, Region is always lowercase, but VersionLabel is all uppercase. Do we care about it being case-sensitive in the first place?

ejaffee01: > Do we care about it being case-sensitive in the first place?

While we could allow both, I feel like it may be better to pick one and stick to it. 

I've updated region to be capitalized, and force it lowercase when generating the request url.

mtian725: Perhaps we should check with Chris first about this.

mtian725:
Other than the playwright tests failing, LGTM

src/Region.ts
ejaffee01: Discussed with Chris offline, he said we should allow both for Env and Region and I've updated the code to reflect this. For VersionLabel, since it's related to search I don't want to allow lowercase just in case this is not compatible with what they use it for, in addition to search not actually using the SDK atm.

mtian725 (approved):

PR #127, opened 2024-01-26
Update documentation

swong1213:
Update README.md to `import { analytics } from "@yext/analytics"` instead of `import { Analytics }...`, and remove usages of `new`.

J=FUS-6200
TEST=manual

Since the test site does `import { analytics } from "@yext/analytics"`, and calls the analytics function without using the `new` keyword, I ran the test site and confirmed that clicking the buttons successfully fired analytics events.


ejaffee01 (approved):
LGTM

PR #130, opened 2024-02-13
EventsSDK: Add debug mode

ejaffee01:
Add a `debug` flag to AnalyticsConfig. When set to true events will not be sent to the Events API and additional information will be logged to console. The added logs are: 

- The AnalyticsConfig following the construction. 
- The Payload following the call to with
- The two payloads that are passed into the merge function
- The final request body that would be sent to the Events API 
   - Will specify if fetch or beacon would be used
   - Will specify request url would be used
   - Will specify if session tracking is enabled
   - Explains to disable debug flag in order to actually fire the event

J=FUS-6239
R=abenno, mtian
TEST=manual

Enabled the debug mode in the test site. See screenshots below for logging examples.
Screenshot 2024-02-13 at 2 17 50 PM
Screenshot 2024-02-13 at 2 20 52 PM


ejaffee01:
Coverage check is failing, but I think that may be okay for this commit. The coverage is still > 80%, and there is not really unit tests to write for the additions made in this PR. We could test what is expected to will print out in the logs, but I think that may be overkill.

etc/analytics.api.md
```diff
@@ -14,6 +14,7 @@ export function analytics(config: AnalyticsConfig): AnalyticsEventService;
 export interface AnalyticsConfig {
     authorization: string;
     authorizationType: 'apiKey' | 'bearer';
+    debug?: boolean;
```
mtian725, line 17: Do we want to add something to the README about this?

src/AnalyticsEventReporter.ts
```diff
@@ -66,6 +83,20 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
...
+          requestUrl +
+          '\n' +
+          'To send the event to the Yext Events API, disable the debug flag in your AnalyticsConfig.\n',
+        finalPayload
+      );
+    }
+
     // If useBeacon returns true, return boolean response of postWithBeacon as string.
     if (shouldUseBeacon) {
       return new Promise((resolve, reject) => {
```
mtian725, line 102: I know we discussed to do this in two steps: add logging to the sdk and drop debug mode requests in PixelGateway, but I think we should still not send any requests if it's in debug mode.

src/AnalyticsEventReporter.ts
```diff
@@ -34,10 +34,27 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
   with(payload: EventPayload): AnalyticsEventService {
     const currentPayload =
       this.payload === undefined ? payload : merge(this.payload, payload);
+    if (this.config.debug) {
+      console.log('[DEBUG] AnalyticsConfig Object:\n', this.config);
```
mtian725, line 38: Try using template literals. This might help with the formatting and readability of the strings

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

src/AnalyticsEventReporter.ts
ejaffee01: Yeah that's my bad I meant to put the post calls behind the debug flag.

etc/analytics.api.md
ejaffee01: It sounds like there are more documentation updates that are needed, maybe we could couple this with those? Iirc you had a list of items for Chris to update in docs, could you add this to that list?

ejaffee01:
Updated screenshots with logs in Chrome (fetch) with session tracking disabled and firefox (beacon) with session tracking enabled: 
Screenshot 2024-02-14 at 11 39 31 AM
Screenshot 2024-02-14 at 11 40 49 AM

src/AnalyticsEventReporter.ts
```diff
@@ -34,10 +34,31 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
...
   }
 
   public async report(newPayload?: EventPayload): Promise<string> {
+    if (this.config.debug) {
+      console.log(
+        `[DEBUG] Merging the following payloads:
+
+        Original Payload from call to \`with\`:
+
+        ${this.payload ? JSON.stringify(this.payload) : '{}'}
```
mtian725, line 54: Is there a reason why this is formatted differently than the other logs? My understanding was to keep this as one line, but can't just just do two console logs?

src/AnalyticsEventReporter.ts
```diff
@@ -34,10 +34,31 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
   with(payload: EventPayload): AnalyticsEventService {
     const currentPayload =
       this.payload === undefined ? payload : merge(this.payload, payload);
+    if (this.config.debug) {
+      console.log('[DEBUG] AnalyticsConfig Object:\n', this.config);
+      console.log(
+        '[DEBUG] Payload following call to `with`:\n',
```
mtian725, line 40: Not sure how big of a difference it is, but probably should say `with()` instead of `with`. Might be more semantically obvious that `with` is referring to the function.

src/AnalyticsEventReporter.ts
```diff
@@ -34,10 +34,31 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
...
+        `[DEBUG] Merging the following payloads:
+
+        Original Payload from call to \`with\`:
+
+        ${this.payload ? JSON.stringify(this.payload) : '{}'}
+
+        New Payload from call to \`report:\`
+        
+        ${newPayload ? JSON.stringify(newPayload) : '{}'}`
+      );
```
mtian725, line 59: Looking over the screenshots you provided, I don't think this is necessary. You're already providing an "after" log below right before the function would be sending a request. 

I think this would result in noise since printing `newPayload` does not provide anything since the user can see what it is in their code. `this.payload` before the merge might be redundant as it will match any previous call to `with()`

src/AnalyticsEventReporter.ts
```diff
@@ -34,10 +34,31 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
   with(payload: EventPayload): AnalyticsEventService {
     const currentPayload =
       this.payload === undefined ? payload : merge(this.payload, payload);
+    if (this.config.debug) {
+      console.log('[DEBUG] AnalyticsConfig Object:\n', this.config);
```
mtian725, line 38: I would move this up to the constructor since you dont have to call `with()` to make a request. When you move this to the constructor, I would also log the EventPayload as well

mtian725 (commented):
Left some comments

src/AnalyticsEventReporter.ts
ejaffee01: Agreed

src/AnalyticsEventReporter.ts
ejaffee01: I had this in the constructor at first, but then this ended up getting logged multiple times before I even clicked a button on the test-site. I think this is because `with` calls the constructor again. You are right that the constructor could be used without `with`, so I am going to move it back to the constructor, but do you think the below seems confusing or is it fine? 

Screenshot 2024-02-14 at 5 06 05 PM

etc/analytics.api.md
ejaffee01: Done.

ejaffee01:
Updated screenshots
Screenshot 2024-02-14 at 5 17 01 PM

Screenshot 2024-02-14 at 5 16 17 PM


Screenshot 2024-02-14 at 5 17 17 PM

src/AnalyticsEventReporter.ts
mkilpatrick: I would say that the logging should only show up when report is called.

src/AnalyticsEventReporter.ts
```diff
@@ -34,10 +34,27 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
...
+        '[DEBUG] Payload following call to `with`:\n',
+        currentPayload
+      );
+    }
     return new AnalyticsEventReporter(this.config, currentPayload);
   }
 
   public async report(newPayload?: EventPayload): Promise<string> {
+    if (this.config.debug) {
+      console.log(
```
mkilpatrick, line 49: I feel like the original payload vs the newPlayload doesn't matter. Just print what would have been sent in the request.

src/AnalyticsEventReporter.ts
```diff
@@ -66,6 +83,20 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
...
+        '[DEBUG] The following Payload would be sent to the Yext Events API using ' +
+          (shouldUseBeacon ? 'Beacon. ' : 'fetch(). ') +
+          'Session Tracking is ' +
+          (this.config.sessionTrackingEnabled ? 'enabled.\n' : 'disabled.\n') +
+          'Request URL: ' +
+          requestUrl +
+          '\n' +
+          'To send the event to the Yext Events API, disable the debug flag in your AnalyticsConfig.\n',
```
mkilpatrick, line 95: This sentence seems unnecessary. Someone should know if they've enabled debug mode or not.

src/AnalyticsEventReporter.ts
```diff
@@ -66,6 +83,20 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
...
+        finalPayload
+      );
+    }
+
```
AjayBenno, line 99: Wait does it still make the request? Do we need to return on line 97?

src/AnalyticsEventReporter.ts
```diff
@@ -66,6 +83,20 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
     const shouldUseBeacon = useBeacon(finalPayload, this.config.forceFetch);
     const requestUrl = setupRequestUrl(this.config.env, this.config.region);
 
+    if (this.config.debug) {
+      console.log(
+        '[DEBUG] The following Payload would be sent to the Yext Events API using ' +
+          (shouldUseBeacon ? 'Beacon. ' : 'fetch(). ') +
+          'Session Tracking is ' +
+          (this.config.sessionTrackingEnabled ? 'enabled.\n' : 'disabled.\n') +
+          'Request URL: ' +
```
AjayBenno, line 92: Is there value in showing the request URL to whoever is using this? Or more value in just showing the config parameters(so they know that they config they set is actually applied).

AjayBenno: also thoughts on just putting the debug mode logic into the post helper?

Would clean up this method and also make sure that we're not duplicating logic/ the debug could be incorrect(i could see a scenario where we change lines 101 onwards but forget to update the debug log line which is just confusing.

src/AnalyticsEventReporter.ts
```diff
@@ -34,10 +34,27 @@ export class AnalyticsEventReporter implements AnalyticsEventService {
...
     return new AnalyticsEventReporter(this.config, currentPayload);
   }
 
   public async report(newPayload?: EventPayload): Promise<string> {
+    if (this.config.debug) {
+      console.log(
+        '[DEBUG] Merging the following payloads:\n' +
+          '\nOriginal Payload from call to `with`:\n' +
+          `\n${this.payload ? JSON.stringify(this.payload) : '{}'}\n` +
+          '\nNew Payload from call to `report:`\n' +
```
AjayBenno, line 53: Thoughts on simple helper for formatting debug messages? Mainly to clean up the readability of this report function + file.

Also helps so that all debug messages have consistent formatting!

src/AnalyticsEventReporter.ts
ejaffee01: Should be resolved by moving debug logic to the post methods

ejaffee01:
Updated screenshot: 
Screenshot 2024-02-15 at 11 52 36 AM

mkilpatrick (approved):

PR #132, opened 2024-02-20
EventsSDK: Document eligible $auto fields

ejaffee01:
The SDK EventPayload does not make it clear if certain fields do not need to be set by the user because they can be automatically inferred when the the value is set to `$auto` (which is the default value for several fields).

This change adds documentation to the EventPayload so these fields are indicated. 

[J=FUS-6251](https://yexttest.atlassian.net/browse/FUS-6251)
R=abenno, mtian
TEST=manual

Ran `npm run build` to confirm everything is formatted correctly.


src/EventPayload.ts
```diff
@@ -71,22 +80,48 @@ export interface EventPayload {
   internalUser?: boolean;
   /** The IP address for the event.*/
   ip?: {
-    /** The IPv4 address associated with the event. */
+    /** The IPv4 address associated with the event.
+     *
+     * Defaults to '$auto' which indicates to use the Source IP address from the IP header of the request.
+     * This IP address is only used if the `Yext-Opt-In: ip` header is present."
```
AjayBenno, line 86: This IP address is only used if the `Yext-Opt-In: ip` header is present. 

Is this actually true? I didn't think it was.

AjayBenno (changes requested):

src/EventPayload.ts
ejaffee01: I copied this from the Events Schema description, so we'd want to update that too if this isn't the case.

src/EventPayload.ts
```diff
@@ -12,9 +12,18 @@ export interface EventPayload {
   action?: Action;
   /** The authorization token for the request. This will be setup from the Key or Bearer in the config. */
   authorization?: string;
-  /** Whether the event is the result of bot activity. */
+  /** Whether the event is the result of bot activity.
+   *
+   * Defaults to '$auto' which indicates to
```
mtian725, line 17: I think this phrasing is a bit confusing because to someone not to familiar with the overall system `What is '$auto'? I'm only seeing that this could be a boolean, but what type is '$auto'? If it's a string, I can imagine that this conflicting information is confusing.`

If we are just going to update the documentation, then I would avoid referencing '$auto' altogether.

mtian725: I would also figure out a consistent phrase or template to use. 

Maybe we just use something like `Will default to <some value> when processed if omitted`

src/EventPayload.ts
ejaffee01: Updated the schema in alpha.

src/EventPayload.ts
```diff
@@ -118,7 +147,11 @@ export interface EventPayload {
     /* The ID of the template from which a site was generated. */
     template?: string;
   };
-  /** The timestamp at which the event occurred, in ISO format. */
+  /** The timestamp at which the event occurred, in ISO format.
```
mtian725, line 150: Might be useful to provide a template or example

src/EventPayload.ts
```diff
@@ -71,22 +79,43 @@ export interface EventPayload {
...
+   * If not specified the system will automatically
+   * determine all location information from the request's IP address,
+   * or the value of the `ip` property if specified.
+   */
   location?: Coordinates | string;
-  /** The URL of the page where the event occurred */
+  /** The URL of the page where the event occurred.
+   *
+   * If not specified the system will automatically
+   * use the `Referer` header of the request.
```
mtian725, line 111: I would also specific here that the `Referer` header is automatically added when the request is made and is not something the user needs to provide

mtian725: Perhaps "....use the `Referer` header from the auto-generated request headers". Also, comment applies to the other changes that mentioned headers as well.

src/EventPayload.ts
```diff
@@ -109,22 +147,21 @@ export interface EventPayload {
...
-    scope?: string;
-    /* The UID of the site an event was tied to. */
-    siteUid?: number;
-    /* The ID of the template from which a site was generated. */
-    template?: string;
-  };
-  /** The timestamp at which the event occurred, in ISO format. */
+  /** The timestamp at which the event occurred, in ISO format.
+   * For example, September 27, 2022 at 6 p.m. is represented as
+   * 2022-09-27 18:00:00.000.
```
mtian725, line 152: I forget. Do we convert to UTC or is what is passed in presumed to be in UTC? If it is the latter, we probably should mention that. If not, this is fine.

src/EventPayload.ts
```diff
@@ -71,22 +79,52 @@ export interface EventPayload {
...
+  /** The locale of the user who generated the event.
+   *
+   * If not specified the system will automatically
+   * determine the locale from the Accept-Language header.
+   */
   locale?: string;
   /** The location information of the visitor for the event.
    * Either a Coordinates object with both latitude and longitude or a string
-   * with the country of the visitor for the event, as a ISO 3166-1 alpha-2 country code. */
+   * with the country of the visitor for the event, as a ISO 3166-1 alpha-2 country code.
```
mtian725, line 101: Would it be too much to provide a link to the set of ISO 3166-1 alpha-2 codes? Looking at the existing codes, most common countries are generally straightforward, but might be useful to provide a direct link to look it up. Wonder what your opinion on this is? Might need to bounce this off with Chris.

src/EventPayload.ts
```diff
@@ -109,22 +147,21 @@ export interface EventPayload {
...
+   * If not specified the system will automatically
+   * use the time when the request was received as the event timestamp.
+   */
   timestamp?: Date | string;
   /** The monetary value of the event. */
   value?: {
     /** The monetary value. */
     amount: number;
-    /** The ISO 4217 currency code of the currency the value is expressed in. */
+    /** The ISO 4217 currency code of the currency the value is expressed in.
```
mtian725, line 162: Also here, would it be too much to provide a link to the set of code?

src/EventPayload.ts
```diff
@@ -71,22 +79,52 @@ export interface EventPayload {
...
     /** The algorithm to use to anonymize the IP address after collection. */
     algorithm: string;
   };
   /** A label assigned to the event, e.g. a CTA label. */
   label?: string;
-  /** The locale of the user who generated the event. */
+  /** The locale of the user who generated the event.
+   *
+   * If not specified the system will automatically
+   * determine the locale from the Accept-Language header.
```
mtian725, line 96: auto-generated Accept-Language header

src/EventPayload.ts
```diff
@@ -71,22 +79,52 @@ export interface EventPayload {
   internalUser?: boolean;
   /** The IP address for the event.*/
   ip?: {
-    /** The IPv4 address associated with the event. */
+    /** The IPv4 address associated with the event.
+     *
+     * If not specified the system will use the Source IP address
+     * from the IP header of the request.
```
mtian725, line 85: auto-generated  IP header

src/EventPayload.ts
ejaffee01: Discussed here https://yext.slack.com/archives/G01SSAYJVM4/p1709063238616539

src/EventPayload.ts
ejaffee01: Discussed here https://yext.slack.com/archives/G01SSAYJVM4/p1709063238616539

mtian725 (approved):

PR #133, opened 2024-02-20
Analytics SDK: Update region/environment to be only upper case

pcammarata1:
Instead of allowing upper and lower case region and environment names, allow just upper-case names.

J=FUS-6250
TEST=manual

Ran npm i/npm run build/npm run test and made sure that all unit tests passed.


ejaffee01 (changes requested):
Left a comment. Also you'll want to put the reviewers under `Reviewers` instead of `Assignee` so we get notified correctly.

ejaffee01 (approved):
LGTM

AjayBenno (approved):

PR #135, opened 2024-02-20
AnalyticsSDK: Add new optional scope Event Payload field for pages

pcammarata1:
This CR creates a new optional string field that denotes the scope of pages actions in event payloads.

J=FUS-6253
TEST=none

### DIFF
--- a/src/EventPayload.ts
+++ b/src/EventPayload.ts
@@ -111,6 +111,8 @@ export interface EventPayload {
   sessionId?: string | null;
   /** Fields specific to reporting Yext Pages Analytics Events */
   pages?: {
+    /** The scope of an action. Applies to Pages events. */
+    scope?: string;
     /* The UID of the site an event was tied to. */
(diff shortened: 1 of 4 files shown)


src/EventPayload.ts
```diff
@@ -111,6 +111,8 @@ export interface EventPayload {
   sessionId?: string | null;
   /** Fields specific to reporting Yext Pages Analytics Events */
   pages?: {
+    /** The scope of an action. Applies to Pages events. */
```
AjayBenno, line 114: Are we holding off adding tests using this until we implement the backend?

AjayBenno (approved):

src/EventPayload.ts
pcammarata1: That's what I was thinking, but do you think we can go ahead and add tests?

AjayBenno (approved):

PR #137, opened 2024-02-27
Support Beacon fallback for stable version

AjayBenno:
Support falling back to beacon for the current stable version. If the browser is detected to be firefox, send the event via beacon instead of fetch w/keepalive.

Also bumped version to v0.6.6.

### DIFF
--- a/src/infra/ChatAnalyticsReporter.ts
+++ b/src/infra/ChatAnalyticsReporter.ts
@@ -57,7 +57,6 @@ export class ChatAnalyticsReporter {
    */
   public async report(event: ChatEventPayLoad): Promise<EventAPIResponse> {
     const headers: Record<string, string> = {
-      Authorization: `KEY ${this.apiKey}`,
       'Content-Type': 'application/json',
     };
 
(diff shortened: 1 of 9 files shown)


src/infra/ChatAnalyticsReporter.ts
```diff
@@ -68,7 +67,8 @@ export class ChatAnalyticsReporter {
       this.endpoint,
       {
         ...event,
-        sessionId
+        sessionId,
+        authorization: `KEY ${this.apiKey}`
```
ejaffee01, line 71: Out of curiosity, why is it preferable to pass the key via request body vs in header?

ejaffee01 (approved):
LGTM!

src/infra/ChatAnalyticsReporter.ts
AjayBenno: Beacon doesn't support auth headers which is why we needed to move this to be sent via the request body. Figured its simpler to just do it for both cases instead of conditionally for beacon.

PR #138, opened 2024-02-27
EventsSDK: Add originalEventName to EventPayload

ejaffee01:
Add originalEventName to EventPayload to support Pages. 

[J=FUS-6279](https://yexttest.atlassian.net/browse/FUS-6279)
R=abenno, mtian
TEST=manual

Tried passing pages object with originalEventName value into test site and confirmed no compile time errors.


src/EventPayload.ts
```diff
@@ -113,6 +113,8 @@ export interface EventPayload {
   pages?: {
     /** The scope of an action. Applies to Pages events. */
     scope?: string;
+    /** The legacy identifier of the event. */
```
AjayBenno, line 116: let's not use legacy as it indicates some connotation with the event.

ejaffee01: Ack

mtian725 (approved):

PR #144, opened 2025-02-04
analytics: upgrade github actions

Fondryext:
This PR upgrades various github actions from their deprecated v3 versions to v4.


mtian725 (approved):
Thanks for doing this

PR #147, opened 2025-02-13
EventsSDK: make entityId optional for search's CtaEvent

anguyen-yext2:
In search, a click event fired on a Generative Direct Answer card might not have an entity associated with it. Specifically, if the direct answer text has a link, we do not know which entity that link comes from, but we want to be able to report the link click event as a CTA event.

### DIFF
--- a/docs/analytics.ctaevent.entityid.md
+++ b/docs/analytics.ctaevent.entityid.md
@@ -9,5 +9,5 @@ The entity ID for the entity.
 <b>Signature:</b>
 
 ```typescript
-entityId: string;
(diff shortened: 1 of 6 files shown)


ddseo:
Change LGTM but I believe we want to merge changes directly into main, rather than to a release branch.

anguyen-yext2:
> Change LGTM but I believe we want to merge changes directly into main, rather than to a release branch.

I followed the same procedure I did [here](https://github.com/yext/analytics/pull/146) to release a new version of the legacy SDK.

PR #152, opened 2025-04-08
EventsSDK: update pageUrl and referrerUrl defaults

brleeYext:
Changed the defaults for the pageUrl and referrerUrl so they no longer use the Referrer header as a default.
Also increased SDK version to 1.0.3

https://yexttest.atlassian.net/browse/FUS-7183


ejaffee01:
Changes LGTM but 2 things
- Do you think this should be a minor version upgrade or patch version? This feels like new functionality over a bug fix, but might be worth confirming with the team. 
- Any easy way to update/add a test for this change? If not simple no worries

ejaffee01 (commented):
Changes LGTM but 2 things

Do you think this should be a minor version upgrade or patch version? This feels like new functionality over a bug fix, but might be worth confirming with the team.
Any easy way to update/add a test for this change? If not simple no worries

brleeYext:
> Do you think this should be a minor version upgrade or patch version?

Good catch, it should be a minor version as we are changing the defaults for `pageUrl` and `referrerUrl`, adding backwards compatibility for `pageUrl` and removing the past default for `referrerUrl`. Added tests and updated the version in new commit.

ejaffee01 (approved):

